PR #14, opened 2018-02-02
Extend capability of standings.py

kunalcsc630:
Added ability to pull standings up to 1903 instead of up to 1969. Working on extending ability all the way to 1871.

### DIFF
--- a/pybaseball/standings.py
+++ b/pybaseball/standings.py
@@ -61,6 +61,7 @@ def standings(season=None):
         raise ValueError("This query currently only returns standings until the 1903 season. Try looking at years from 1903 to present.")
     # retrieve html from baseball reference
     soup = get_soup(season)
+    table = None
     if season>=1969:
         tables = get_tables(soup, season)
     elif season>=1903:
--- a/pybaseball/standings.py
+++ b/pybaseball/standings.py
@@ -1,4 +1,5 @@
 from bs4 import BeautifulSoup
+from bs4 import Comment
 import requests
 import datetime
 import pandas as pd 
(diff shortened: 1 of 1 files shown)


jldbc:
Love the idea here. I'm guessing this is meant to return league-level standings for seasons before 1969? 

Running locally I get an error: 
```
from standings import standings
test = standings(1944)
  File "/Users/jledoux/Documents/projects/pybaseball/pybaseball/standings.py", line 71, in standings
    tables = get_tables(code, season)
  File "/Users/jledoux/Documents/projects/pybaseball/pybaseball/standings.py", line 31, in get_tables
    headings = [th.get_text() for th in table.find("tr").find_all("th")]
UnboundLocalError: local variable 'table' referenced before assignment
```

kunalcsc630:
Yep. This is bizarre because when I run it locally, it works perfectly fine. I've added one statement. Maybe that's why you're getting errors?
screen shot 2018-02-02 at 9 09 36 am
screen shot 2018-02-02 at 9 10 14 am
screen shot 2018-02-02 at 9 10 30 am

jldbc:
Still getting the same error on my end. It looks like the issue is coming from line 31:
`headings = [th.get_text() for th in table.find("tr").find_all("th")]`

`table` isn't yet defined here (it isn't passed from `standings` to `get_tables`, so setting it before calling the function on line 64 won't help). 

For post-1969 `table` exists in `headings = [th.get_text() for th in table.find("tr").find_all("th")]` because it's within the loop `for table in tables:`. I think changing the name of line 30's variable from `tables` to `table` solves this, since pre-1969 isn't iterating through a list of tables the way its post-1969 counterpart is.

kunalcsc630:
Ah yes! I now notice that in my screenshot, I use the variable `tables` but on GitHub, the variable is `table`. I believe that should fix the issue. Does it work for you now?

kunalcsc630:
Just extended capability to standings from 1871 onwards, the earliest possible year from which you can pull standings.

jldbc:
Your fix to the previous problem works now, but pre-1903 queries error out for the same reason as before: 

UnboundLocalError: local variable 'tables' referenced before assignment

Did you mean to change line 71 to `else:` instead of `elif season>=1903:` so that a pre-1903 query is processed by the block of code within that else statement? Otherwise, `tables` won't exist on line 77 if `season` < 1903.

Once this is fixed these changes should be ready to merge.

kunalcsc630:
Yeah. Hopefully everything else works after that fix.

jldbc:
Looks good now. Thanks!

PR #15, opened 2018-02-02
Minor fixes to pitching_leaders.py

ntanners:
I made two changes:
1) I fixed a bug in the hard-coded list of stats.  Previously, FB% was pulled twice: once as 49 and once as 75.  In addition, IFH% was not pulled.  Because the headings were hard-coded, this caused some headings to be mis-applied: FB% was labeled IFFB%, IFFB% was labeled HR/FB, and HR/FB was labeled IFH%.  I removed 49 from the list of stats and added 52.
2) I changed the Headings list so that it is drawn from the fangraphs output, rather than from a hard-coded list.  This allowed me to catch the bug described above.

Thanks for making this awesome package, and I hope this edit is helpful.  It's my first pull request!


jldbc:
Good catch! This function's batting equivalent might have the same issue if this was wrong.

ntanners:
Thanks!  Last night I checked the headings list in batting_leaders.py, and
I didn't find any errors there, but when I have time, I'll try to make the
same change where the code pulls the list of headings from the fangraphs
output.

I was also wondering if it might be possible to pull projection data from
fangraphs, but I haven't figured out how to construct a url that gets all
of the projection data on a single page.  Adding "&page=1_100000" to the
end of the url doesn't seem to work for the projections page like it does
for the leaderboards page.

Is there somewhere one can look for information about how to make custom
urls for fangraphs?  Or is it the case that fangraphs does not want people
to write programs to download their projection data?

Thanks,

Nadav

PR #16, opened 2018-02-07
Minor edits to batting_leaders

ntanners:
I made the same change to batting_leaders that I previously made to pitching_leaders, namely pulling the stats headings from fangraphs, rather than hard-coding the headings.  This change results in two separate columns having the heading 'FB%', which created an error.  To fix that error, I renamed the second occurrence of 'FB%' to 'FB% (Pitch)', which is the same value that it had previously been assigned.

I also made a minor change to the code that creates the dataframe to make it a bit more succinct.


jldbc:
Sorry for the delay, but looks good to me. Thanks!

PR #17, opened 2018-02-16
Few fixes

ghost:
If you call the schedule_and_record function on 2/16/18 for 2018 season, originally it would try to get the link to the 2018 season, which we can guarantee won't be made until April when the MLB season begins. I fixed that bug.


kunalcsc630:
I've added more to `playerid_lookup`. Now, it allows for spelling mistakes in users' searches.

PR #21, opened 2018-04-03
Added 'pitch_name' to not_numeric list

WillAyd:
Cool library! Messing with the stat cast data I came across a bug that should be fixed with this PR


jldbc:
Thanks! Looks like this column is breaking things today, I'll push this out soon.

Issue #22, opened 2018-04-26
Splitting Statcast Requests into Multiple Queries


andersonfrailey:
Continuing a discussion that was started in issue #20, I think one of the ways we can solve the issues users are having when they try to run `statcast_pitcher` and `statcast_batter` queries over periods longer than about two months would be to break those queries up.

My general approach to this would be to first determine if a query is longer than some arbitrary maximum (probably around 60 days), then use some of the features in Python's datetime package to iterate over the user specified period in chunks calling the necessary function each time. This will result in a list of DataFrames returned by each function call and which can be bound together.

A secondary goal would be to run the queries in parallel, but I think that can wait until after this initial work is done.

I'm happy to work on adding this feature and would love to hear any feedback/ideas from others.

jldbc:
Closing this out since the PR was merged. 

Your point on running these in parallel might not make a huge difference for these two functions, but it's definitely something I've considered for the `statcast` function where common queries require a larger number of requests. That might be a worthy project for another day.

Thanks!

Issue #20, opened 2018-04-03
Baseball Savant Scraping Issue


rensmacneill:
Hi there, 

I first want to say thank you for your awesome work in putting together pybaseball. I was playing around with the package yesterday and ran the fivethirtyeight New Science of Hitting example with no issues. Today I was running some statcast and statcast_batter queries in much the same way as in the example, however I run into errors every time.

I now receive the following error when attempting to print the shape of the initial dataset in the New Science of hitting example: ValueError: could not convert string to float: 'Sinker'.

When running statcast_batter with an end-date as in the example I receive an Error: Query Timeout message. However when statcast_batter is run without an end-date, it returns without an error, however the player_id is ignored in the returned data set. 

I have attached an image to better illustrate what I'm experiencing (run here with David Ortiz's player_id as in the statcast_batter example).
![screen shot 2018-04-03 at 10 39 26 pm](https://user-images.githubusercontent.com/7392763/38249829-5a9679d4-3790-11e8-8c49-4d7905e34121.png)

I have run this using both Python 2.7.10 and Python 3.6.1, with the same behaviour for both.

EDIT: I have tried manually scraping the data using the technique set out by @alanrkessler and baseball savant is returning a 'Error: Query Timeout. Please try to limit your query to less data.' as the singular field of the csv when you attempt to scrape the data.

jldbc:
Thanks for flagging, I'll look into this. In the other `statcast` function this timeout issue is caused by requesting more than a certain number of rows, which could be fixed by breaking it into multiple requests and iterating. I could see Ortiz hitting the max row count given his career length. 

The 'Sinker' ValueError sounds like a column may have changed.

rensmacneill:
According to Darren Willman and Bill Petti on twitter (see [https://twitter.com/BillPetti/status/981155132585988096](url) ) there are some additional fields in each baseballsavant query result now, which would explain why the ValueError is appearing.

I'll mess around with iterating through larger requests tonight. Could the extra fields be causing the player_id to be ignored in a one-day query?

Thanks for your response also @alanrkessler, I'll let you know how things go.

jldbc:
**For statcast_batter (and by extension probably statcast_pitcher)**
[...]

PR #23, opened 2018-04-30
Split Statcast Queries

andersonfrailey:
This PR is an implementation of the solution to issue #20 I proposed in issue #22. So far I've only implemented this change in `statcast_batter` and `statcast_pitcher`, I haven't tried anything with the rest of the `statcast` functions. I created a new `utils.py` file to avoid some code duplication.

I've run `statcast_batter` and `statcast_pitcher` a few times and everything seems to work as it should. Labeled WIP because I want to play with the other `statcast` functions to see if they have any of the same request timeout errors before saying it's ready for a full review.

### DIFF
--- a/pybaseball/__init__.py
+++ b/pybaseball/__init__.py
@@ -1,3 +1,4 @@
+import pybaseball.utils
 from .playerid_lookup import playerid_reverse_lookup
 from .playerid_lookup import playerid_lookup
 from .statcast import statcast, statcast_single_game
--- a/pybaseball/statcast_batter.py
+++ b/pybaseball/statcast_batter.py
@@ -1,52 +1,18 @@
-import pandas as pd
-import requests
-import datetime
-import warnings
-import io
+from pybaseball.utils import sanitize_input, split_request
 
-def validate_datestring(date_text):
-    try:
-        datetime.datetime.strptime(date_text, '%Y-%m-%d')
-    except ValueError:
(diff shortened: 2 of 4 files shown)


andersonfrailey:
Looks like the `statcast` function already has a method for breaking up queries so I'll leave it be in this PR and remove the WIP label.

jldbc:
Sorry, lost track of this. Looks good to me though at first glance, the move to a utils file is also much needed. Is there any reason behind picking 60 days as the time interval to move by and not something larger? My guess is we could afford a larger step size which would make this run faster

andersonfrailey:
@jldbc, I picked 60 days because I found that once I started trying queries for time intervals larger than about 2 months the requests would time out. I can experiment a little and see if there are slightly larger intervals I could use.

jldbc:
Are you using the bugfix from 6d0952c51bc3f00a4ff54b6a137a0b82eafe2d75? The size of these requests is a lot smaller now that it's filtering for players instead of returning everything. I got it to work with 365+ days as the interval. Let me know what you find after testing it some and I'll merge whatever's working best.

andersonfrailey:
@jldbc, I tested it out a bit after the bug fix you pointed out and could run queries up to six years before running into a timeout issue. I'll update this PR accordingly.

jldbc:
Looks good. Most players won't hit the 40k event mark on this yet since we only have this data 2008 - present, but this should future-proof this as we start to get more data

PR #26, opened 2018-06-23
Complete some schedule and results to-do's

andersonfrailey:
This PR takes care of a couple of the to-do's in `team_results.py`. I've capitalized the team abbreviation and raised an error when season is greater than the current year.

I started to take work on raising an error when the season is before the team's first season and adding the capabilities to return results for every team, but I realized it was more complicated than I originally thought it would be. My original idea was to just include a dictionary that contained each team's abbreviation and their start year, but there are a ton of historical teams that can currently be found with `schedule_and_records` so that dictionary would need to be huge to cover everyone. I think it would be better to wait and see if there's a better way to handle it.

### DIFF
--- a/pybaseball/team_results.py
+++ b/pybaseball/team_results.py
@@ -3,8 +3,8 @@ import pandas as pd
 import requests
 from bs4 import BeautifulSoup
 from datetime import datetime
+from pybaseball.lahman import teams
 
-# TODO: raise error if year < first year of a team's existence
 # TODO: retrieve data for all teams? a full season's worth of results
 
 def get_soup(season, team):
@@ -95,9 +95,23 @@ def make_numeric(data):
 
 def schedule_and_record(season=None, team=None):
     # retrieve html from baseball reference
+    # sanatize input
     team = team.upper()
+    df = teams()
+    seasons = df['yearID']
+    team_names = df['teamIDBR']
+    first_occurrences = team_names.drop_duplicates(keep='first')
+    first_occurrences_index = first_occurrences.index.values
+    first_years = pd.DataFrame(seasons[first_occurrences_index])
+    first_years['teamID'] = first_occurrences
+    first_season_map = dict(zip(first_years['teamID'],
+                                first_years['yearID']))
+    if season < first_season_map[team]:
+        m = "Season cannot be before first year of a team's existence"
+        raise ValueError(m)
     if season > datetime.now().year:
         raise ValueError('Season cannon be after current year')
+
     soup = get_soup(season, team)
     table = get_table(soup, team)
     table = process_win_streak(table)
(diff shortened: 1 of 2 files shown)


jldbc:
A dictionary is the best approach that comes to my mind here, but Lahman can probably help with this. 

Will this dict work?

```
import pandas as pd
from pybaseball.lahman import teams
df = teams()
seasons = df['yearID']
team_names = df['teamIDBR']
first_occurrences = team_names.drop_duplicates(keep='first')
first_occurrence_index = first_occurrences.index.values
first_years = pd.DataFrame(seasons[first_occurrence_index])
first_years['teamID'] = first_occurrences
first_season_map = dict(zip(first_years.teamID, first_years.yearID))
```

andersonfrailey:
@jldbc I'll try out your suggestion this weekend and let you know how it works out!

andersonfrailey:
@jldbc your suggestion works great. Implemented in latest commit.

pybaseball/team_results.py
```diff
@@ -95,8 +95,25 @@ def make_numeric(data):
...
+    first_occurrences_index = first_occurrences.index.values
+    first_years = pd.DataFrame(seasons[first_occurrences_index])
+    first_years['teamID'] = first_occurrences
+    first_season_map = dict(zip(first_years['teamID'],
+                                first_years['yearID']))
+    if season < first_season_map[team]:
+        m = "Season cannot be before first year of a team's existence"
+        raise ValueError(m)
+    if season > datetime.now().year:
+        raise ValueError('Season cannon be after current year')
```
jldbc: nit: cannot*

pybaseball/team_results.py
```diff
@@ -95,8 +95,25 @@ def make_numeric(data):
...
+    team = team.upper()
+    df = teams()
+    seasons = df['yearID']
+    team_names = df['teamIDBR']
+    first_occurrences = team_names.drop_duplicates(keep='first')
+    first_occurrences_index = first_occurrences.index.values
+    first_years = pd.DataFrame(seasons[first_occurrences_index])
+    first_years['teamID'] = first_occurrences
+    first_season_map = dict(zip(first_years['teamID'],
+                                first_years['yearID']))
```
jldbc: Thoughts on hardcoding this instead of generating it each time the function is called? Re-downloading Lahman's teams file at each function call will be slow

pybaseball/team_results.py
```diff
@@ -95,8 +95,25 @@ def make_numeric(data):
...
+    if season < first_season_map[team]:
+        m = "Season cannot be before first year of a team's existence"
```
jldbc: If the dict from line 107 is hardcoded, this logic should be modified to allow requests to succeed if a team isn't in the dict (e.g. expansion franchises and team relocations). Otherwise `first_season_map[team]` will throw an error if an unknown `team` is passed

andersonfrailey:
Added a hard coded dictionary. The dictionary wasn't actually that large and it reduced run time.

jldbc:
LGTM!

PR #27, opened 2018-06-24
Add Retrosheet functions

andersonfrailey:
This PR adds some functions to pull game logs from Retrosheet. I included some documentation and the notice required by Retrosheet to use their data. I ran all of the functions and everything seemed to work as expected. Let me know if there's anything else I need to do/add.


andersonfrailey:
@jldbc like the play-by-play ones? I haven't yet, but was going to try and work on that soon. Could probably make some progress this week/weekend.

jldbc:
Let's handle that under a new PR since this one looks ready to merge. The event files have an odd file type so I don't want to slow this one down unnecessarily in case they end up being complicated.

PR #31, opened 2018-09-26
Suggestion: switch parser to lxml -- I think it should give a noticeable speedup

schorrm:
My testing suggests a roughly 33% reduction in CPU time necessary, just by switching to the lxml parser. Really minimal code changes, just adding lxml to dependencies and "lxml" instead of "html.parser".

### DIFF
--- a/docs/retrosheet.md
+++ b/docs/retrosheet.md
@@ -1,0 +1,36 @@
+# Retrosheet
+
+The functions in `retrosheet.py` retrieve game log data from [Retrosheet.org](http://www.retrosheet.org/gamelogs/index.html).
+
+## Retrosheet Data Notice
+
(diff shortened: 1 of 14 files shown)


jldbc:
Good stuff! I will run some speed tests over the weekend.

schorrm:
results in?

gjreda:
@jldbc Related to this PR, would you be open to a similar PR that abstracts the various `get_soup` functions into a more generic one? My thought was to have `get_soup(url)` which returns the page's content as a BS object, along with having a dictionary of `URL_MAPPINGS` in either `utils.py` or a new `constants.py`.

This would hopefully follow DRY principles and allow for easier testing in the future.

I'd be happy to contribute this change if you're interested.

PR #32, opened 2018-09-26
lahman fixes

bentrevett:
- no longer have to re-download the lahman db every time you want to perform a lahman function

- download_lahman only needs to be called once, get_lahman_zip will return None if lahman DB already exists in current working directory

### DIFF
--- a/pybaseball/lahman.py
+++ b/pybaseball/lahman.py
@@ -9,10 +9,13 @@ url = "http://seanlahman.com/files/database/baseballdatabank-2017.1.zip"
 base_string = os.path.join("baseballdatabank-2017.1","core")
 
 def get_lahman_zip():
-	# Retrieve the Lahman database zip file. Making this a function since everything else
-	# will be re-using these lines
-	s=requests.get(url,stream=True)
-	z = zipfile.ZipFile(BytesIO(s.content))
+	# Retrieve the Lahman database zip file, returns None if file already exists in cwd. 
+	# Making this a function since everything else will be re-using these lines
+	if os.path.exists(base_string):
+		z = None
+	else:
+		s = requests.get(url, stream=True)
+		z = zipfile.ZipFile(BytesIO(s.content))
 	return z
 
 def download_lahman():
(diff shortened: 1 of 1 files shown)


jldbc:
Thanks for submitting this. I'm noticing two issues:

First: if I call `download_lahman()` twice, it throws an error because the second time it's calling `z.extractall()` on `None`. Since it's expected to have z = None if this is the second time the function has been called, I think it's fine to just add an if statement to only make this call if z is not None. 

Second: if I call any of these functions without having called `download_lahman` first, it runs at the same speed as before, since `download_lahman` is the only function that actually writes the database to the working directory. So, if I call `download_lahman`, after that everything is blazing fast. But if I call `parks` twice in a row, it's just as slow the second time. 

Tbh I'm not sure what the best solution to # 2 is. The easiest is to have any lahman function download the full db if it's not already in the directory, so every call after that is fast. But I also don't love the idea of forcing an unwanted download on people. Maybe the best solution is to fix # 1 and leave # 2 as-is. Thoughts?

bentrevett:
Sorry about that first error, should be fixed now. 

As for # 2: one messy solution is making each function take an optional `download` argument which will ensure `download_lahman()` is called at least once.

```python
def parks(download=False):
    # do this for every table in the lahman db so they can exist as separate functions
    if download:
        download_lahman()
    z = get_lahman_zip()
    f = os.path.join(base_string, "Parks.csv")
    data = pd.read_csv(f if z is None else z.open(f), header=0, sep=',', quotechar="'")
    return data
```

Or could possibly add a global `download` variable the user can set, e.g.

```python
download = False

def parks():
    # do this for every table in the lahman db so they can exist as separate functions
    if download:
        download_lahman()
    z = get_lahman_zip()
    f = os.path.join(base_string, "Parks.csv")
    data = pd.read_csv(f if z is None else z.open(f), header=0, sep=',', quotechar="'")
    return data
```

And then if a user wants to set `download` to `True` by default, they can do something like:

``` python
import pybaseball 

pybaseball.lahman.download = True
```

Let me know what you think.

schorrm:
I think I have a better idea for this.
Let's rewrite get_lahman_zip to this:
`
_ziphandle = None
def get_lahman_zip():
	global _ziphandle
	# Retrieve the Lahman database zip file, returns None if file already exists in cwd. 
	# Making this a function since everything else will be re-using these lines
	if os.path.exists(base_string):
		_ziphandle = None
	else:
		if not _ziphandle:			
			s = requests.get(url, stream=True)
			_ziphandle = zipfile.ZipFile(BytesIO(s.content))
	return _ziphandle`

This way: if it's downloaded, no bother, and if it isn't, well, we have session persistence and a major speedup for subsequent runs

schorrm:
Okay, here's an idea that would work with the unzipped downloaded database form. It's a little wacky, but here it is:
```
class _zipHandler():
	def __init__(self):
		if os.path.exists(base_string):
			self.local = True
		else:
			self.local = False
			s = requests.get(url, stream=True)
			self.handle = zipfile.ZipFile(BytesIO(s.content))
	
	def open (self, path):
		if self.local:
			return path
		else:
			return self.handle.open(path)

_zipinstance = None
def get_lahman_zip():
	# Retrieve the Lahman database zip file, returns None if file already exists in cwd. 
	# Making this a function since everything else will be re-using these lines
	global _zipinstance
	if not _zipinstance:
		_zipinstance = _zipHandler()
	return _zipinstance
```
This would allow things like:
```
def all_star_full():
	z = get_lahman_zip()
	f = os.path.join(base_string, "AllstarFull.csv")
	data = pd.read_csv(z.open(f), header=0, sep=',', quotechar="'")
	return data
```
to work as is -- the handler would plug in either the file string or the file as appropriate. it's just require a different `download_lahman()` that wouldn't use `get_lahman_zip()`, something like this:
```
def download_lahman():
	if os.path.exists(base_string):
		return None
	else:
		s = requests.get(url, stream=True)
		z = zipfile.ZipFile(BytesIO(s.content))
		z.extractall()
```

schorrm:
My apologies for the stupid comments - which I deleted.
In much simpler form -- we can keep all of Ben's changes, and just swap the `get_lahman_zip()` part for the following:
```
_handle = None
def get_lahman_zip():
	# Retrieve the Lahman database zip file, returns None if file already exists in cwd. 
	# Making this a function since everything else will be re-using these lines
    global _handle
	if os.path.exists(base_string):
		_handle = None
	else:
        if not _handle:
            s = requests.get(url, stream=True)
            _handle = zipfile.ZipFile(BytesIO(s.content))
	return _handle
```
It's only a couple lines difference, and keeping the `_handle` around gives us session persistence. So if we have a call to download() then we get full local performance, otherwise, we can keep the zip around throughout the duration.

jldbc:
Closing this out after #41 was merged

PR #34, opened 2018-10-10
Same as Ben's Lahman fix, just with persistent session zipfile for when it isn't downloaded.

schorrm:
Basically, if you call, say, parks() w/o dling, it'll keep the ZipFile there in a global scoped variable for persistence.
Then download_lahman will take the ZipFile if it is exists, download it otherwise, and extract.
So the following:
```python
p = parks() # will trigger putting the ZipFile in local memory
b = batting() # will use the already existing ZipFile in memory
download_lahman() # will extract the already existing ZipFile, and clear the dled one
b = batting() # will use the already extracted dbase
```
I think it gives the ideal balance between speed and not forcing downloading more stuff / clutter. It's only about 8MB of memory, so not a big deal imho.
Thanks


schorrm:
Can we take a look at this?

schorrm:
bump?

jldbc:
Sorry, took a break from this for a bit. These changes look good to me. Remove the other stuff that wasn't part of this issue (that can be a different pr) and I'll merge

schorrm:
Oops. Sorry about that. I'll fix that now. I'll open a new PR to clean up the history, it's already pretty bad here (sorry about that :( ), if that's okay?

jldbc:
no problem

PR #41, opened 2019-02-24
Revised Lahman fixes

schorrm:
Several things. Almost same code as previous PR, just without the previous garbage, and with the following additional changes:

1. I used the Chadwick Databank URL on GitHub, it's better hosting, I think, and also stays up to date.
2. In the newest one, with stats from 2018, they renamed `Master.csv` to `People.csv`. I accounted for the new filename in the code, added a `people()` function to preserve the symmetry, though I left in `master` as an alias to `people` to preserve backwards compatibility.


jldbc:
Good call on moving this to the Chadwick source, that should be more reliable between seasons. LGTM!

PR #42, opened 2019-02-25
Added module for Collecting Player Salary Data from Spotrac

TAThor:
Added the file 'player_salaries.py' to collect player salary data from Spotrac. Also added documentation for the newly added functions.


TAThor:
This is accidentally branched off team_batting_ref rather than master, and I am not sure how to fix this. Is this a problem?

Sorry for the mistake.

PR #43, opened 2019-02-26
Added 'player_salaries.py' to collect player salary data from Spotrac

TAThor:
Module collects player salaries from Spotrac.

P.S. I had previously opened a pull request for this same feature, but had accidentally branched it from 'team_batting_ref' rather than 'master'. I went ahead and fixed that.

### DIFF
--- a/docs/player_salaries.md
+++ b/docs/player_salaries.md
@@ -1,0 +1,79 @@
+# Player Salaries from Spotrac
+
+Note: Salary Data on Spotrac is only available beginning in 2011.
+
+`salaries_by_all(start_season, end_season=None, to_float=False)`
+
+PROBLEM: requests to receive all player salaries only returns the top 100 salaries, and doesn't have pages to iterate through. Not sure how to workaround yet.
+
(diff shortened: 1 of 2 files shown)


jldbc:
Can you provide some detail on how this data differs from the salary data in the Lahman database? I'm not too familiar with Spotrac salary data

TAThor:
> 
> 
> Can you provide some detail on how this data differs from the salary data in the Lahman database? I'm not too familiar with Spotrac salary data

I'm not very familiar with the Lahman database, but having just downloaded it there appears to be some differences:

1. The current Lahman Database (downloaded from [here](http://www.seanlahman.com/baseball-archive/statistics/)] only has salary data up to 2016. Spotrac data is available through 2024 for contracts that extend into that season.

2. The Lahman Database has salary data going back to 1985 while Spotrac only goes back to 2011.

3. The Lahman Database contains columns for team and league, where 'player_salaries' currently does not. This can be fixed in a somewhat hacky way. I'll look into doing this soon.

schorrm:
Can you please add this PR to my fork?

PR #35, opened 2018-10-31
Added the ability to search on a list of players in playerid_lookup. …

timandrews335:
I added the ability to supply a list of strings to search for in playerid_lookup as noted in TODO.  I also adjusted the documentation to provide an example.

PR #49, opened 2019-03-21
Added 'batch_playerid_lookup' to 'playerid_lookup.py' to allow playerid lookups for multiple players.

TAThor:
While doing analysis on multiple players I realized a new lookup table had to be generated for each player, and this could take a significant amount of time. So, I added this functionality.

I noticed that this was in the TODO's listed atop 'playerid_lookup.py', so it resolves that.


timandrews335:
You had the same ideas as me.  I'm waiting for my similar pull request to be merged (#35)

Issue #58, opened 2019-06-23
Question - Statcast pitcher spin rate


joshhejka:
When using statcast_pitcher, there is a column "spin_rate_deprecated". There is also "spin_dir", "break_angle_deprecated", and "break_length_deprecated", all of which appear to be "NaN" for 2019 stats at the very least. Is there any way to access spin rate/spin direction/etc. for 2019 since these columns are now deprecated?

tpoatsy3:
In short, no, there is no way to access the old data. In fact, I can't even get the data from 2018 or earlier. However, I did find a paper from Professor Nathan from the University of Illinois on how to calculate the spin axis from other statcast data. I'm going to add a function that calculates the spin axis and magnitude to compensate for their deprecation. Could you post whatever old data you might have so that I can be sure that the values I end up with are similar to what you might have (orientation and units are my biggest concerns).

jldbc:
The oldest data I have (in terms of time collected) is from Dec 2017, and these were all null by that point in time. I think your best bets here are either to find an old pitchfx csv in someone else's repo, or to see if Prof Nathan has any reproducible examples you can use to verify that your approach is working

tpoatsy3:
Thanks! In the meantime, I created a `statcast_pitcher_spin` function that shows the user how much break on each pitch is due to spin and two angles of the spin axis. These seemed to be the most interested calculations. They're all defined in the documentation I wrote up for the function. I'm creating a pull request for my contribution now.

schorrm:
Closed by #64

PR #64, opened 2019-08-30
Statcast pitcher spin rate fix

tpoatsy3:
This merge adds the `statcast_pitcher_spin` function to the library. The new function piggy-backs on the existing `statcast_pitcher` function, but adds calculations of the pitch's movement as a result of its spin and two angles measuring the spin's axis of rotation. The physics and formulae that I used are based on the work of Prof. Alan Nathan of the University of Illinois.

I used the test driven development method, so I also included my test scripts and data. This will validate the calculations and could provide a home for scripts that test other parts of the library.

Lastly, I included documentation for the function that mirrors the rest of the package's docs. The examples have a bias towards pitchers from the Chicago Cubs, but everything else should be in order.


tpoatsy3:
@jldbc I wanted to follow up on this to see if there are any changes that you think should be made before it's merged into pybaseball

tpoatsy3:
@jldbc Just circling back on this. This pull should close issue #58. Let me know if you think other changes should be made or if you have any concerns with the code that are preventing making the pull.

schorrm:
@tpoatsy3 This is really good -- can you please make this PR to my fork? We're trying to get this going again

schorrm:
Okay, I'm trying to merge the PR, but I can't get the unit testing to work here.
Also, I don't *love* the setup where it goes through a module indirection

schorrm:
@tpoatsy3 -- can you please drop the gitignore from this, merge up to date?
and can you take a look at the __init__ thing? Can we get this a different way?

schorrm:
Also, in terms of structure, I think this might be better as a boolean flag in the function calls for statcast. And possibly even on by default.

schorrm:
@tpoatsy3, you still there?

schorrm:
@bdilday can you take a look at this? @TheCleric started working on fixing the unit tests and there are discrepancies here apparently

TheCleric:
@bdilday I checked my fixes on a branch on my copy of the repo:

https://github.com/TheCleric/pybaseball/tree/statcast-fix

The main issue is the integration test. It seems like the numbers in the example CSV and the ones being generated are a few hundredths off. Not sure if it's an algorithm issue, or that the CSV just needs to be updated.

bdilday:
I did some digging on this. I was able to find the spreadsheet that the algorithm is based on here, http://baseball.physics.illinois.edu/trackman/MovementSpinEfficiencyTemplate.xlsx

I can say that the algorithm implemented in this PR is definitely consistent with Alan Nathan's excel-based computations. The example data for [this test](https://github.com/jldbc/pybaseball/pull/64/files#diff-0b58f37bc0505464d8be1c5cbeb93748d5277526c1def63172b099942c3f3420R48) is actually from the spreadsheet, so we've effectively got a test for consistency, not only for the final result but for each individual component. 

On the other hand if I take the the data from the `live_Darvish_July` [data ](https://github.com/jldbc/pybaseball/pull/64/files#diff-da94b2280934026f8909b34248712ec896145ae1e8de077315a40edd641245c1) included in this PR, and put it into the spreadsheet, the answer is identical to @tpoatsy3 code and is not identical to the values in the csv file itself. i.e., the csv file appears to be wrong (if we assume Alan Nathan's calculations to be correct). 

it's hard to know why these aren;t consistent - presumably this test passed at some point? so that would suggest the data being returned by mlbam has changed. we also don't know where the Darvish data came from (@tpoatsy3 ?) or if we should expect it to match Alan Nathan's spreadsheet.  

overall, we don't need to block on the test of the Darvish data passing. It may be worth doing the pull from mlbam (`df = spin.statcast_pitcher_spin(start_dt='2019-07-01', end_dt='2019-07-31', player_id=506433)` , computing the spin stuff, saving that file, and testing against that going forward. This would allow for a regression test based on changing mlbam data.

TheCleric:
Thanks @bdilday ! I'll update the CSV file then with the current data and we'll go from there.

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,140 @@
...
+
+K = .005153 #Environmental Constant
+SIG_DIG = 4
+
+def statcast_pitcher_spin(start_dt=None, end_dt=None, player_id=None):
+	pitcher_data = statcast_pitcher(start_dt, end_dt, player_id)
+
+	spin_df = pitcher_data[['release_extension', 'vx0', 'vy0', 'vz0', 'ax', 'ay', 'az', 'release_spin_rate']].copy()
+
+	spin_df = find_release_point(spin_df)
```
bdilday, line 13: I think it'd be helpful if these calculations were in a separate function, so they'd be easier to test

something like, 

```
spin_df = find_intermediate_values(spin_df)
pitcher_data[['Mx', 'Mz', 'phi', 'theta']] = spin_df[['Mx', 'Mz', 'phi', 'theta']].copy()
return pitcher_data

def find_intermediate_values(spin_df):
    spin_df = find_release_point(spin_df)
    spin_df = find_release_time(spin_df)
    .
    .
    return spin_df
```

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,140 @@
+from pybaseball import statcast_pitcher
+import pandas as pd
+import numpy as np
+
+K = .005153 #Environmental Constant
```
bdilday, line 5: would it be useful to compute this instead of hard-coding it? i.e. assume 70 F sea-level as a default but let a user specify something different in the call to `statcast_pitcher_spin`?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,140 @@
...
+SIG_DIG = 4
```
bdilday, line 6: is rounding really necessary? if so maybe the result should be rounded only in the last step?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,140 @@
...
+# def get_statcast_pither_test_data():
+# 	df = pd.read_csv("tests/statcast_pitching_test_data.csv")
+# 	return df
+
+def find_release_point(df):
+	df['yR'] = (60.5 - df['release_extension']).round(SIG_DIG)
+	return df
+
+def find_release_time(df):
+	df['tR'] = time_duration(df['yR'], df['vy0'], df['ay'], 50, False).round(SIG_DIG)
```
bdilday, line 42: should 50 be variable at top of the file?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,140 @@
...
+		360 + np.arctan2(df['amagz'], df['amagx'])*180/np.pi)+90
+	df['phi'] = df['phi'].round(0).astype('int64')
+	return df
+
+def find_lift_coefficient(df):
+	df['Cl'] = (df['amag']/(K*df['vbar']**2)).round(SIG_DIG)
+	return df
+
+def find_spin_factor(df):
+	df['S'] = (0.4*df['Cl']/(1-2.32*df['Cl'])).round(SIG_DIG)
```
bdilday, line 100: a doc string with a reference to where this formula came from would be helpful

pybaseball/statcast_pitcher_spin.py
TheCleric: As well, wouldn't this be park dependent? I.e., I'm assuming Tropicana Field (sea-level) vs Coors Field (a mile up) would possibly be different.

TheCleric:
@schorrm I've updated my branch with @bdilday's suggestions and all tests pass now. Since the original submitted seems to have disappeared, we can work off that branch if you'd like.

bdilday:
I don't see this feature as being urgent, maybe we could send him on email to let me know we want to merge this and give him the opportunity to finish this off? It's available on linked in profile, https://github.com/tpoatsy3

tpoatsy3:
I didn't see all of this activity on this thread. I'll review your comments this week and make some changes. Thanks for the interest in adding this.

pybaseball/statcast_pitcher_spin.py
tpoatsy3: It's great that you brought this up because I was hoping on replacing K with a calculation later. The reason I didnt in the original PR was to manage scope and keep the test data could be consistent.

pybaseball/statcast_pitcher_spin.py
tpoatsy3: I had also thought that rounding at each step was superfluous, but I compared my testing data to the excel sheet that I was doing the calculations for and there were marginal errors that compounded as I went further in the calculations and amounted to non-insignificant differences in the output.

If we can verify the accuracy of the functions I built, then we can absolutely remove rounding at every step

pybaseball/statcast_pitcher_spin.py
tpoatsy3: yeah, that's a good call, I'll make that change too

pybaseball/statcast_pitcher_spin.py
schorrm: I would make a suggestion that we leave out the rounding at each step perhaps for now and then ask Alan Nathan nicely on Twitter if he can check if the math is accurate?

schorrm:
Also @tpoatsy3 -- take a look at @TheCleric's work on the testing for this -- it may be of use

pybaseball/statcast_pitcher_spin.py
bdilday: I think that;s unnecessary. I confirmed that this code agrees with Alan Nathan's spreadsheet, not just the end result but each individual cell.

schorrm:
Hey, @tpoatsy3  -- any progress?

schorrm:
@tpoatsy3 -- anything new doing on this?

schorrm:
Anyone? Anyone? Bueller?

tpoatsy3:
@schorrm I'm making these edits now. I'll get something pushed later and hopefully we can resolve this.

pybaseball/statcast_pitcher_spin.py
tpoatsy3: Both of these numbers are actually explained in the README tab of Prof. Nathan's excel workbook. Y Velocity is captured at Y=50 and the final data points are taken at y=17/12. These numbers will be defined at the top of the file as such.

tpoatsy3:
Just as a heads up, Prof Nathan changed the function for `S`. The latest push includes that change.

schorrm:
@tpoatsy3 -- can you please drop the .gitignore from your PR to resolve that conflict?

pybaseball/statcast_pitcher_spin.py
schorrm: Bumping this again -- @bdilday -- should we round intermediate steps or only at the end in the final formatted df?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -1,140 +1,200 @@
...
-	df['vxbar'] = (2*df['vxR'] + df['ax']*df['tf'])/2
-	df['vybar'] = (2*df['vyR'] + df['ay']*df['tf'])/2
-	df['vzbar'] = (2*df['vzR'] + df['az']*df['tf'])/2
+    df['vxbar'] = (2*df['vxR'] + df['ax']*df['tf'])/2
+    df['vybar'] = (2*df['vyR'] + df['ay']*df['tf'])/2
+    df['vzbar'] = (2*df['vzR'] + df['az']*df['tf'])/2
+
+    df['vxbar'] = df['vxbar'].round(4)
+    df['vybar'] = df['vybar'].round(4)
+    df['vzbar'] = df['vzbar'].round(4)
```
schorrm, line 102: I'm not sure that we want to be rounding, but if we are, this should be `SIG_DIG`

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,200 @@
...
+    return df
+
+
+def find_average_velocity(df):
+    df['vbar'] = three_comp_average(df['vxbar'], df['vybar'], df['vzbar']).round(SIG_DIG)
+    return df
+
+
+def find_average_drag(df):
+    df['adrag'] = (-(df['ax']*df['vxbar'] + df['ay']*df['vybar'] + (df['az'] + 32.174)*df['vzbar'])/ df['vbar']).round(SIG_DIG)
```
schorrm, line 112: should be `GRAVITATIONAL_ACCELERATION`, no?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,200 @@
...
+    return df
+
+
+def find_lift_coefficient(df):
+    df['Cl'] = (df['amag']/(K*df['vbar']**2)).round(SIG_DIG)
+    return df
+
+
+def find_spin_factor(df):
+    df['S'] = (0.166*np.log(0.336/(0.336-df['Cl']))).round(SIG_DIG)
```
schorrm, line 150: Can we comment / explain this line here?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,200 @@
...
+    return df
+
+
+# HELPERS
+def special_round(series, digit):
+    series = series * 10**digit
+    series = np.where(series >= 0, series + .5, series - .5)
+    series = series.astype('int64').astype('float64')
+    series = series / 10**digit
+    return series
```
schorrm, line 180: Can this get a comment / explanation?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def time_duration(s, v, acc, adj, forward):
+    """
+        Finds flight time given an original position, velocity, accelaration, and target position.
+
+        Direction does not affect the time duration. It helps assign a positive or negative
+        value to the flight time.
+
+        s = (pd.Series) spacial point at known time
```
schorrm, line 190: *spatial

pybaseball/statcast_pitcher_spin.py
schorrm: OK yeah then nerf the roundign

schorrm:
Looks good -- can we nerf the rounding?

tpoatsy3:
yeah, I'll make a last commit removing it and we can call it closed. (FYI, getting rid of the rounding might cause some unit test failures)

schorrm:
Can you re-run the thingy and just update the tests accordingly?

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+			equal
+		"""
+
+		comp_df = pd.DataFrame()
+		comp_df['left'] = df1[name].round(SIG_DIG)
+		comp_df['right'] = df2[name].round(SIG_DIG)
+		comp_df['diff'] = comp_df['left'] - comp_df['right']
+		comp_df['diff'] = comp_df['diff'].abs().round(SIG_DIG)
+		# print(comp_df.query('diff > 0.0000'))
+		return comp_df.query('diff > .0001').empty
```
TheCleric, line 33: It would probably be simpler to use pandas `assert_frame_equal` with the `check_exact` and `atol` parameters:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+
+SIG_DIG = 4 # Required because of read_csv bug in pandas
+
+rounding_error_columns = ['vxR', 'vyR', 'vxbar', 'vybar', 'vbar']
+
+class TestStatcastPitchingSpinCalcs(unittest.TestCase):
+
+	# Helper Methods
+	def compare_columns(self, df1, df2, name):
+		return df1[name].equals(df2[name])
```
TheCleric, line 15: Probably safer to use `assert_series_equal` here (though I do not have a lot of experience with it):

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_series_equal.html

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+			'find_theta': ['theta'],
+		}
+
+		for method, columns in test_dict.items():
+			func = getattr(spin, method)
+			test_frame = func(test_frame)
+
+
+			for column in columns:
+				logging.info("Begin testing on {}".format(column))
```
TheCleric, line 73: f strings are preferred over `format` strings.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+						self.assertTrue(self.compare_almost_equal(test_frame, target_frame, column))
+
+					else:
+						self.assertTrue(self.compare_columns(test_frame, target_frame, column))
+
+					logging.info("{} passed".format(column))
+
+
+				except Exception:
+					logging.exception("Tests on {} have failed".format(column))
```
TheCleric, line 87: f strings are preferred over `format` strings.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+
+				try:
+					if column in rounding_error_columns:
+					# Almost equal assertion is necessary for small differences that arise after consecutive calculations
+						self.assertTrue(self.compare_almost_equal(test_frame, target_frame, column))
+
+					else:
+						self.assertTrue(self.compare_columns(test_frame, target_frame, column))
+
+					logging.info("{} passed".format(column))
```
TheCleric, line 83: f strings are preferred over `format` strings.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+			'find_magnus_magnitude': ['Mx', 'Mz'],
+			'find_phi': ['phi'],
+			'find_lift_coefficient': ['Cl'],
+			'find_spin_factor': ['S'],
+			'find_transverse_spin': ['spinT'],
+			'find_spin_efficiency': ['spin eff'],
+			'find_theta': ['theta'],
+		}
+
+		for method, columns in test_dict.items():
```
TheCleric, line 67: Instead of this I think it would be better to use `pytest.mark.parametrize` to feed these into a simpler function. See `tests/integration/pybaseball/analysis/projections/marcels/test_marcels_batting.py` for a simple example.

tests/test_statcast_pitching_spin.py
TheCleric: Also if this is moved to a `parametrize` style test you won't need the `try...except` block, which would be preferred in a test.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+
+			This example is of Yu Darvish(506433) from 2019-07-01 to 2019-07-31
+			Darvish has a wide variety of pitches, giving the most diverse set of test data
+
+		"""
+		# Import the rubric
+		template_data = pd.read_csv('tests/statcast_spin/live_Darvish_July2019_test.csv').round(SIG_DIG)
+
+		# Run the method in question
+		df = spin.statcast_pitcher_spin(start_dt='2019-07-01', end_dt='2019-07-31', player_id=506433)
```
TheCleric, line 107: If this is pulling real data (which it appears it is) then it should be somewhere in the `tests/integration` folder.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+
+		# Columns needed to be checked
+		target_columns = ['Mx', 'Mz', 'phi', 'theta']
+
+		for column in target_columns:
+			logging.info("Begin testing on {}".format(column))
+
+
+			if column in ['Mz']:
```
TheCleric, line 116: Why loop through 4 columns, and then ignore 3 of them?

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+		template_data = pd.read_csv('tests/statcast_spin/live_Darvish_July2019_test.csv').round(SIG_DIG)
+
+		# Run the method in question
+		df = spin.statcast_pitcher_spin(start_dt='2019-07-01', end_dt='2019-07-31', player_id=506433)
+
+		# Columns needed to be checked
+		target_columns = ['Mx', 'Mz', 'phi', 'theta']
+
+		for column in target_columns:
+			logging.info("Begin testing on {}".format(column))
```
TheCleric, line 113: f strings are preferred over format strings.

tests/test_statcast_pitching_spin.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+			if column in ['Mz']:
+			# Almost equal assertion is necessary for small differences that arise after consecutive calculations
+				self.assertTrue(self.compare_almost_equal(df, template_data, column))
+
+			else:
+				self.assertTrue(self.compare_columns(df, template_data, column))
+
+			logging.info("{} passed".format(df, template_data, column))
```
TheCleric, line 123: f strings are preferred over format strings.

tests/test_statcast_pitching_spin.py
TheCleric: Also if this is done that way, I would just eliminate this function and do the `assert_series_equal` directly in your tests.

TheCleric (changes requested):

TheCleric:
Now that the master branch's tests have been fixed, please merge that in here as well to clear up some of the broken tests. :)

tpoatsy3:
Thanks for the feedback. I'll start tying things together here.

tpoatsy3:
Just checking in, I rebased my branch on the master branch, but I'm still working on getting testing to pass

schorrm:
This doesn't seem like a unit testing bug -- is it possible that your reference data is out of date?

schorrm:
I'm guessing you need to regen because of solving the cumulative rounding issue

bdilday:
I think the csv file is just outdated, not related to rounding. I did some digging on this [here](https://github.com/jldbc/pybaseball/pull/64#issuecomment-710078795)

TheCleric:
As well I noticed that the structure of where the files are does not follow the current standard. I completely get why though as when you began this I don't even know if we had any/many unit tests! :)

If you could though, please put all unit tests in the `tests/pybaseball` directory mirroring the package structure (e.g. since `statcast_pitcher_spin.py` is in the project root, `test_statcast_pitcher_spin.py` would go in `tests/pybaseball/` directly). As well any unit test `.csv` files should go in `tests/pybaseball/data`.

Thanks and sorry to make this PR even longer. Just want to make sure we remain consistent. :)

tpoatsy3:
yup, thanks. I didnt mean to push my code that last time after I had rebased my branch on master. I'm in the process of re-writing some of the tests using the pandas' testing library and refreshing the data since some of the calculations changed. One area of difficulty is that, I'm running into problems getting tests to pass after the spin_factor calculation. I took a few days away from this, so I'll take another look with fresh eyes to see if I can find what's driving the tests to fail.

cro1100:
Hi, I'm interested in getting involved with this project.  I'm a relative
newbie to the world of data engineering, but I think I have a general sense
of what is available at this point.  What I don't understand is quite how
the code all fits together.  I've searched the documentation and haven't
been able to find something which can give me a more complete sense of what
I'm looking at.  Can someone contact me to get the ball (pun intended)
rolling?  I'm most curious about adding the vertical and horizontal
movement of pitchers' pitches.

Thanks!

Charles Ro

schorrm:
@tpoatsy3 -- can you see if regenerating like @bdilday suggested above fixes that -- I trust Ben's judgment here.

tpoatsy3:
I've tried regenerating the files with the new formulas, but once we take out the aggressive rounding, I cant get the first test to pass (`test_individual_calculations`). Specifically, it fails on `find_lift_coefficient`. I think it's being caused by the difference in how python and excel are treating the `K` value (which is a constant for air density). I was looking into this and apparently MS Excel has its own proprietary decimal rounding which departs from IEEE standards. Since all of the previous calculations work before, it must be how Excel is handling the decimal number compared to python.

Just to be clear, before I refreshed the data, I changed the `Cl` calculation in the excel data file from being updatable to being a static `0.005383' in the excel doc, which matches the K global variable that I have in the file. I let the excel doc reload and then saved that output as a csv. The numbers largely match and when I use the `check_less_precise=3` attribute with `pandas._testing.assert_frame_equal` function, I'm getting a difference of 6.47%. With check_less_precise=2, it's 0.04%.

Is there another way to test this data that's not circular without using excel? I'm open to ideas.

bdilday:
@tpoatsy3 could you push your changes?I think it'd be easiet to try to untangle this if we could see the most up-to-date code. Alos, can you clarify which excel file you're using, i.e. what's the link on Alan Nathan's website?

pybaseball/statcast_pitcher_spin.py
```diff
@@ -0,0 +1,216 @@
...
+
+Article: http://baseball.physics.illinois.edu/trackman/SpinAxis.pdf
+Excel Workbook: http://baseball.physics.illinois.edu/trackman/MovementSpinEfficiencyTemplate-v2.xlsx
+"""
+
+from pybaseball import statcast_pitcher
+import pandas as pd
+import numpy as np
+
+K = .005153  # Environmental Constant
```
bdilday, line 14: this looks like a typo. Should be 0.005383

bdilday:
I believe I've addressed the issues here, with this PR https://github.com/tpoatsy3/pybaseball/pull/2

* hardcodes K = 0.005383 in the excel file

* formats the excel file to avoid rounding in display

* exports a query for Yu Darvish data, using the current statcast values, which have diverged from the test file included in the original PR

tpoatsy3:
@bdilday Thank you so much for the help. I'm not sure how you fixed the excel rounding issue, but it's passing that unittest now.

@schorrm I re-wrote the testing file to use pytest and the utility functions in `tests/pybaseball/conftest.py`. All of the tests from `tests/pybaseball/test_statcast_pitching_spin.py` pass.

schorrm:
Will be going through as soon as the rest of the CI passes... can you pull back up to date?

schorrm:
@TheCleric ?

schorrm:
Upon discussing with @bdilday, I think we'll merge and fix the unrelated FG bugs later.

tpoatsy3:
Ok, sounds good. I'll do a final commit that's rebased on the current master commit (1b8bb70)

tpoatsy3:
I'm all set, feel free to merge whenever

PR #71, opened 2020-05-26
Retrosheet

valdezt:
This is the first of two PRs I'd like to contribute to add retrosheet data to pybaseball. Included in this PR:
- Changed Retrosheet file source to the Chadwick bureau Github. This avoids having to hardcode a max/min date assumption as Retrosheet is always adding data. This does require PyGithub in order to verify that files exist before they're downloaded.
- Added functions to pull schedules, rosters, park codes and event files (no parsing of event files is done yet)
- I bumped the version (though you may want to see that you agree with the new version)
- Added docs for new functions and additional (optional) setup for allowing more PyGithub queries with a CLI access token


schorrm:
This is in here through the fork

PR #76, opened 2020-07-15
migrate batting date range calls to fangraphs

nickwan:
This changes the soup request URL from bbref to fangraphs


schorrm:
Closing for now, as #77

PR #77, opened 2020-07-15
migrate pitching date range calls to fangraphs

nickwan:
This changes the soup request URL from bbref to fangraphs

### DIFF
--- a/pybaseball/league_pitching_stats.py
+++ b/pybaseball/league_pitching_stats.py
@@ -78,7 +78,7 @@ def pitching_stats_range(start_dt=None, end_dt=None):
         raise ValueError("Year must be 2008 or later")
     if datetime.datetime.strptime(end_dt, "%Y-%m-%d").year < 2008:
         raise ValueError("Year must be 2008 or later")
-    # retrieve html from baseball reference
+    # retrieve html from fangraphs 
     soup = get_soup(start_dt, end_dt)
     table = get_table(soup)
     table = table.dropna(how='all')  # drop if all columns are NA
--- a/pybaseball/league_pitching_stats.py
+++ b/pybaseball/league_pitching_stats.py
@@ -2,6 +2,7 @@ import requests
 import pandas as pd
 import numpy as np
 import io
+import warnings
 from bs4 import BeautifulSoup
 import datetime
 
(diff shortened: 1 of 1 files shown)


schorrm:
Nick, can you make these PRs against my fork?

pybaseball/league_pitching_stats.py
```diff
@@ -42,15 +43,16 @@ def get_soup(start_dt, end_dt):
...
         return None
-    url = "http://www.baseball-reference.com/leagues/daily.cgi?user_team=&bust_cache=&type=p&lastndays=7&dates=fromandto&fromandto={}.{}&level=mlb&franch=&stat=&stat_value=0".format(start_dt, end_dt)
+    url = f"https://www.fangraphs.com/leaders.aspx?pos=all&stats=pit&lg=all&qual=1&type=c,4,5,6,7,8,9,10,11,114,12,13,14,15,16,17,18,19,20,21,22,23,24,36,37,38,40,120,121,217,41,42,43,44,117,118,119,45,124,62,122,229,240,251,262,273,230,241,252,263,274,231,242,253,264,275,234,245,256,267,278,226,237,248,259,270,235,246,257,268,279,228,239,250,261,272,227,238,249,260,271,232,243,254,265,276,233,244,255,266,277,236,247,258,269,280,63,64,65,66,67,68,69,70,71,72,73,74,115,116,212,213,214,215,58,59,60,54,55,56,57,29,30,31,292,293,294,295,296,297,298,299&season={start_dt[:4]}&month=0&season1={end_dt[:4]}&ind=0&team=0&rost=0&age=0&filter=&players=0&startdate={start_dt}&enddate={end_dt}&page=1_1500"
     s = requests.get(url).content
     return BeautifulSoup(s, "lxml")
 
 
 def get_table(soup):
-    table = soup.find_all('table')[0]
+    warnings.warn("\nPlease consider supporting FanGraphs\nhttps://plus.fangraphs.com/product/fangraphs-membership/", Warning)
```
schorrm, line 52: Okay I'm not _sure_ about this one

schorrm:
@nickwan , I'm maintaining this here now. Can you explain the need to migrate to FanGraphs?

schorrm:
@nickwan ?

nickwan:
Unsure if you're actually asking (since you've seemingly fixed and closed the issue), but this would move away from bbref which seems to be putting more things under a pay wall.

pybaseball/league_pitching_stats.py
nickwan: Why? They offer a free service and ask for membership. Seems fair if we're scraping to at least promote membership in good faith

schorrm: Yeah, but if so, I would suggest doing this with print, not warnings, which has implications here that are incorrect (that something went wrong or is otherwise liable to break).
And if we're going to have a message, we might as well put in 'Downloading from FanGraphs' before that.

schorrm:
I am seriously asking, since the prior bugs were fixed.

pybaseball/league_pitching_stats.py
nickwan: Well that's the problem isn't it? If this goes behind a pay wall like stuff on bbref, it's going to break? Unless you're getting a tap from MLBAM (and with the current gigantic warnings on FG saying they need support to maintain the site), we can't really trust this source will stay the same. That's not different from bbref scraping, except they've figured out a way to monetize at scale so maybe things stay public (or maybe they start tucking more things behind a pay wall). But with FG it seems as though they want to keep things as open as possible, and this is a painless way to continue keeping their service open. 

Technically speaking, I suppose I don't care about warning vs print but a warning seems more appropriate for the above reasons.

nickwan:
If the previous bugs are fixed then there is no reason to move to FG for now

schorrm:
Ok. So given that the bugs are fixed, should we put a pin in this?

nickwan:
If you want. Alternatively, this could be repurposed to serve as a secondary source if bbref is unsupported again.

schorrm:
Closing for now

PR #84, opened 2020-08-29
Add support for fielding data

TheCleric:
This PR will support the fielding data from FanGraphs and Baseball Reference in a similar way to the pitching and batting data.


schorrm:
Sorry about this little delay, I wanted to get this after the release. We're good to go. Thanks!

PR #88, opened 2020-08-30
fixed large_request in statcast to not skip years if a date before ma…

bgunn34:
shifted to old repository.

see https://github.com/schorrm/pybaseball/pull/18

### DIFF
--- a/pybaseball/statcast.py
+++ b/pybaseball/statcast.py
@@ -61,7 +61,7 @@ def large_request(start_dt,end_dt,d1,d2,step,verbose):
         # if these dates are detected, check if the next season is within the user's query
         # if yes, fast-forward to the next season to avoid empty requests
         # if no, break the loop. all useful data has been pulled.
-        if d.month < 4 and d.day < 15:
+        if (d.month == 3 and d.day < 15) or d.month <= 2:
             print('Skipping offseason dates')
             d1 = d1.replace(month=3,day=15,year=d1.year)
             d = d1 + datetime.timedelta(days=step+1)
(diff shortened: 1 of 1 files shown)


pybaseball/statcast.py
```diff
@@ -61,7 +61,11 @@ def large_request(start_dt,end_dt,d1,d2,step,verbose):
         # if these dates are detected, check if the next season is within the user's query
         # if yes, fast-forward to the next season to avoid empty requests
         # if no, break the loop. all useful data has been pulled.
-        if ((d.month < 4 and d.day < 15) or (d1.month > 10 and d1.day > 14)):
+        if d.month < 4 and d.day < 15:
```
TheCleric, line 64: This would potentially miss some dates. E.g., 2/28.

```suggestion
        if (d.month == 3 and d.day < 15) or d.month <= 2:
```

pybaseball/statcast.py
```diff
@@ -61,7 +61,11 @@ def large_request(start_dt,end_dt,d1,d2,step,verbose):
...
+            print('Skipping offseason dates')
+            d1 = d1.replace(month=3,day=15,year=d1.year)
+            d = d1 + datetime.timedelta(days=step+1)
+        elif d1.month > 10 and d1.day > 14:
```
TheCleric, line 68: Same as above, but in reverse:

```suggestion
        elif (d1.month == 11 and d1.day > 14) or d1.month > 11:
```

schorrm:
Looks great, thanks to both y'all

Issue #59, opened 2019-06-26
Set correct data types in Statcast data


djcunningham0:
Currently all numeric columns in the Statcast data are coerced to a float data type. This happens in the `postprocessing` function in `statcast.py`.

```
numeric_cols = ['release_speed','release_pos_x','release_pos_z','batter','pitcher','zone','hit_location','balls',
                'strikes','game_year','pfx_x','pfx_z','plate_x','plate_z','on_3b','on_2b','on_1b','outs_when_up','inning',
                'hc_x','hc_y','fielder_2','vx0','vy0','vz0','ax','ay','az','sz_top','sz_bot',
                'hit_distance_sc','launch_speed','launch_angle','effective_speed','release_spin_rate','release_extension',
                'game_pk','pitcher.1','fielder_2.1','fielder_3','fielder_4','fielder_5',
                'fielder_6','fielder_7','fielder_8','fielder_9','release_pos_y',
                'estimated_ba_using_speedangle','estimated_woba_using_speedangle','woba_value','woba_denom','babip_value',
                'iso_value','launch_speed_angle','at_bat_number','pitch_number','home_score','away_score','bat_score',
                'fld_score','post_away_score','post_home_score','post_bat_score','post_fld_score']

data[numeric_cols] = data[numeric_cols].astype(float)
```

Many of those numeric columns always contain integer values (balls, strikes, outs_when_up, etc.). These columns should be coerced to an int data type.

Perhaps more importantly, many of the other columns are ID values (batter, pitcher, game_pk, etc.). It would make more sense to coerce these columns to a string data type (or int would also be better than float). 

I would recommend the following changes in the `postprocessing` function:
* Remove 'batter', 'pitcher', and all other ID columns from the numeric_cols list.
* Create a new int_cols list with 'balls', 'strikes', and all other columns that always contain integer values. Then coerce these columns to an int data type.
    + **Edit:** The int data type doesn't allow for nulls. If you're willing to make Pandas >= 0.24.0 a requirement, then you could use the Pandas [nullable integer](https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html) data type. (This data type could also be useful in some other places)

schorrm:
@TAThor , @andersonfrailey -- what do you guys think about requiring this?
Quoting from the NumPy docs: "Changed in version 1.0.0: Now uses pandas.NA as the missing value rather than numpy.nan."
So I think we'd probably want to mandate version 1+?

[...]

PR #91, opened 2020-08-31
Add a FanGraphs generic parser

TheCleric:
In this PR I've added a FanGraps generic table parser. To do this, and ensure it should be backwards compatible, I did the following:

1. Added unit testing to the project
2. Added unit testing for the existing team_fielding code
3. Wrote the FanGraphs parser + unit tests
4. Cut team_fielding over to the new parser
5. Ensure all tests pass

Next steps will be to do the following for any existing FanGraphs function on a one by one basis:
1. Backfill unit tests
2. Replace existing parser
3. Ensure all tests pass


schorrm:
One thing we need to keep in mind is that once we fix all the floats that should be ints (see #59) wouldn't this fail the unit tests? Do we just regen all the unit tests for that PR?

schorrm:
Also, what are the performance impacts of this PR?

TheCleric:
> One thing we need to keep in mind is that once we fix all the floats that should be ints (see #59) wouldn't this fail the unit tests? Do we just regen all the unit tests for that PR?

Yes, that would be my thought. We'd verify it passes before, change to ints, then regenerate the csv file.

TheCleric:
> Also, what are the performance impacts of this PR?

Interesting question. I hadn't approached it from that perspective, but rather from code sharing and maintainability. Still a worthy question though. After some local testing: **About 3x Faster!**

Here's what I did. I copied the old team_fielding.py to my repo as team_fielding_old.py

I then wanted to remove the HTTP access piece (especially since FG would rate limit before we could do some meaningful counts), so I wrote this test bed that injects the html from the static file into the requests library.

```python
from pybaseball.team_fielding import team_fielding as new_tf
from pybaseball.team_fielding_old import team_fielding as old_tf
import os
import requests
import timeit

with open(os.path.join('tests', 'pybaseball', 'data', 'team_fielding.html')) as html:
    mock_content = html.read()


def mock_get(url):
    class MockResult:
        def __init__(self):
            self.content = mock_content

    return MockResult()


requests.get = mock_get


def run_new_tf():
    new_tf(2019)


def run_old_tf():
    old_tf(2019)
```

I then did a series of python timeit runs:

```
pybaseball> python` -m timeit -n 100 -r 5 -s "from team_fielding_timeit import run_new_tf" "run_new_tf()"
100 loops, best of 5: 59.3 msec per loop
pybaseball> python -m timeit -n 100 -r 5 -s "from team_fielding_timeit import run_old_tf" "run_old_tf()"
100 loops, best of 5: 192 msec per loop
```

So we went from an average of 192ms to process the FanGraphs data to a DataFrame, down to 59.3ms.

schorrm:
Auto-CI is a very nice touch. Thank you for including that with the unit tests (the main reason I'd want it, anyway, to get a green check with a PR).
Quick question: what's that Flake8 thingy doing there?

pybaseball/datasources/fangraphs.py
```diff
@@ -0,0 +1,48 @@
+import lxml
+from typing import List
+
+import pandas
```
schorrm, line 4: I feel like `import pandas as pd` is a bit more normative / consistent?

tests/pybaseball/datahelpers/test_postprocessing.py
```diff
@@ -0,0 +1,71 @@
...
+            [
+                ['TBR', '1', '2', '50 %', '8'],
+                ['555', '3', '4', '45%', np.nan]
+            ],
+            columns=['Team', 'Runs', 'Hits', 'CS%', 'HR']
+        ).reset_index(drop=True)
+
+        actual_result = postprocessing.coalesce_nulls(sample_unprocessed_result).reset_index(drop=True)
+
+        pd.testing.assert_frame_equal(actual_result, expected_result)
```
schorrm, line 32: Reviewing the [docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html), should we use `check_dtype=False` to allow for changing the types?

pybaseball/datahelpers/postprocessing.py
```diff
@@ -0,0 +1,41 @@
...
+def columns_except(data: pd.DataFrame, columns: List[str]) -> List[str]:
+    return list(np.setdiff1d(data.columns, columns))
+
+
+def convert_numeric(data: pd.DataFrame, numeric_columns: List[str]):
+    # data.loc[data[numeric_cols] == ''] = None
+    # data[numeric_cols] = data[numeric_cols].astype(float)
+    # Ideally we'd do it the pandas way ^, but it's barfing when some columns have no data
+
+    for col in numeric_columns:
```
schorrm, line 24: Could we insert here a quick check for whether the column contains decimals or percentages, and then use that to determine whether float or integer?

pybaseball/datasources/fangraphs.py
TheCleric: It is. I'll fix it.

tests/pybaseball/datahelpers/test_postprocessing.py
TheCleric: We could in some tests (like this one), but would still want it for testing the numeric postprocessor. My opinion is to lean towards specificity, but would be willing to add it if you thought it was appropriate.

pybaseball/datahelpers/postprocessing.py
TheCleric: I thought about that as well once you mentioned #59 but was worried about performance on larger data sets to scan all values and then attempt to determine if they were all ints or all floats. We could certainly add it if you'd like.

TheCleric:
> Quick question: what's that Flake8 thingy doing there?

Remnant of me doing autoformatting. It shouldn't break anything, but I can remove it if you'd like.

pybaseball/datahelpers/postprocessing.py
schorrm: We shouldn't need to check all of them, we should only need to check the first (nonnull?) instance to determine.
Re performance, I was mainly worried about a potential major regression, I'm not trying to extract every last bit here.

TheCleric: Actually I think I have an idea that would work well here. Let me try some things and I'll update the PR.

schorrm:
Looks _pretty_ good.

schorrm:
Awesome work. I think this is ready to go. Your sign-off too? (I nuked the Flake8 thingy, we can deal with formatting separately)

pybaseball/datahelpers/postprocessing.py
TheCleric: Okay, I added some new code that will just try to determine the type at parse time. I reran my performance metrics, and surprisingly this is even faster!

```
pybaseball> python -m timeit -n 100 -r 5 -s "from team_fielding_timeit import run_new_tf" "run_new_tf()"
100 loops, best of 5: 40.5 msec per loop
```

As well this would likely address the issue in #59 if we inject the try_parse function into the read_csv code, possibly using the converters parameter.

schorrm: wowwwwwwww

schorrm: wowwwwwwwwwww

PR #94, opened 2020-09-02
Team Batting (FanGraphs) => Add unit tests; use new consolidated data source

TheCleric:
This PR is the next step for #91 , now for Team Batting.

In this PR we've added unit tests, and cut team_batting to using the new shared method for getting FanGraphs data.

I've also removed the extraneous parameters in the FanGraphs datasource functions that are now auto-determined.

I've also added a MyPy checker to CI that will NOT fail the PR, but just so we can slowly ensure data types are being respected (could catch potential bugs).


schorrm:
Does this dramatically change the number of columns, though?

schorrm:
Never mind, that's a dumb comment, I got this confused with the regular batting.

pybaseball/datasources/fangraphs.py
```diff
@@ -9,7 +9,7 @@
 ROOT_URL = 'https://www.fangraphs.com'
 
 
-def get_fangraphs_tabular_data_from_html(html: str, xpath: str, percentage_columns: List[str] = None, non_numeric_columns: List[str] = None) -> pd.DataFrame:
+def get_fangraphs_tabular_data_from_html(html: Union[str, bytes], xpath: str) -> pd.DataFrame:
```
schorrm, line 12: <3

pybaseball/team_batting.py
```diff
@@ -1,79 +1,36 @@
...
-            data[col] = data[col].astype(float)/100.
-        else:
-            pass
-
-    # convert columns to numeric
-    not_numeric = ['Team']
-    numeric_cols = [col for col in data.columns if col not in not_numeric]
-    data[numeric_cols] = data[numeric_cols].astype(float)
-    return data
-
```
schorrm, line 56: This is some sweet sweet dividends here

pybaseball/team_fielding.py
```diff
@@ -30,9 +30,7 @@ def team_fielding(start_season, end_season=None, league='all', ind=1):
 
     fg_data = fangraphs.get_fangraphs_tabular_data_from_url(
         _FG_TEAM_FIELDING_URL.format(start_season=start_season, end_season=end_season, league=league, ind=ind),
-        '//table[@class="rgMasterTable"]',
-        percentage_columns=['CS%', 'lgCS%'],
-        non_numeric_columns=['Team', 'Name']
+        '//table[@class="rgMasterTable"]'
```
schorrm, line 33: Substantive: is this always "rgMasterTable"? Can we streamline this call?

tests/pybaseball/datasources/test_fangraphs.py
```diff
@@ -60,9 +60,7 @@ class TestDatasourceFangraphs:
     def test_get_table(self, sample_html, sample_processed_result):
         actual_result = fangraphs.get_fangraphs_tabular_data_from_html(
             sample_html,
-            '//table',
-            ["CS%"],
-            ["Team"]
+            '//table'
```
schorrm, line 63: Is it only ever something else in the testing framework? Can we default it to rg and have the `//table` only in the tests? wondering

pybaseball/team_fielding.py
TheCleric: Great question. Let me check.

tests/pybaseball/datasources/test_fangraphs.py
TheCleric: If so I can just adjust the HTML of the test to add that class.

pybaseball/team_batting.py
TheCleric: I always love deleting code more than adding code.

pybaseball/team_fielding.py
TheCleric: So here is all the places we'd end up implementing this in the future:

- batting_leaders.py
- pitching_leaders.py
- team_pitching.py

All of them use this table class. So I could have this as a default param. I'll modify it shortly.

tests/pybaseball/datasources/test_fangraphs.py
TheCleric: Or not because it's actual FG HTML, so it's already there. lol

TheCleric:
Updated with suggestions.

schorrm:
Looking good. Wonderful work, as usual.

PR #96, opened 2020-09-02
team_pitching to new FanGraphs datasource

TheCleric:


schorrm:
Like clockwork. LGTM. You ready?

tests/pybaseball/conftest.py
```diff
@@ -21,3 +22,19 @@ def get_dataframe(filename: str) -> pd.DataFrame:
         return pd.read_csv(os.path.join(data_dir, filename), index_col=0).reset_index(drop=True)
 
     return get_dataframe
+
+@pytest.fixture()
+def response_get_monkeypatch(monkeypatch):
```
schorrm, line 27: Actually, just the one thing: can you please type annotate this and add a docstring for this function?

TheCleric: You read my mind (or I read your mind?). Just added the annotations. I'll add the docstring.

TheCleric:
Added docstrings

pybaseball/team_batting.py
```diff
@@ -9,7 +9,7 @@
 
 _FG_TEAM_BATTING_URL = "/leaders.aspx?pos=all&stats=bat&lg={league}&qual=0&type=8&season={end_season}&month=0&season1={start_season}&ind={ind}&team=0,ts&rost=0&age=0&filter=&players=0&page=1_100000"
 
-def team_batting(start_season, end_season=None, league='all', ind=1):
+def team_batting(start_season:int, end_season:int=None, league:str='all', ind:int=1):
```
schorrm, line 12: looks a little off

TheCleric: Yeah since I turned off the autoformatter, it's not adding the space I would be accustomed to. lol

schorrm: Something like yapf or something less agressive? Idk. I guess we should really define a proper house style sometime.

TheCleric: I tried yapf, black, and autopep8. Of those autopep8 is the most configurable and least opinionated in my experience. I tend to go with splitting lines > 120 characters and then letting autopep8 do everything else it would like to. But you can configure all the options in a repo wide config file.

TheCleric: If you'd like I can open an issue, show some sample default outputs for a given file from yapf, black, and autopep8, and we can decide if we can tweak what we like/don't like.

schorrm:
LGTM. Thanks a ton, as always.

PR #101, opened 2020-09-03
Batted ball trajectories (pybbda integration)

bdilday:
This PR implements [Alan Nathan's batted ball calculator](http://baseball.physics.illinois.edu/trajectory-calculator-new.html)


pybaseball/datahelpers/postprocessing.py
```diff
@@ -62,3 +63,33 @@ def convert_percentages(data: pd.DataFrame, columns: List[str]) -> pd.DataFrame:
             pass
 
     return data
+
+
+
+
+def check_is_zero_one(instance, attribute, value):
+    if value not in [0, 1]:
+        raise ValueError("{} must be either 0 or 1, not {}".format(attribute, value))
```
schorrm, line 72: `f"{attribute} must be either 0 or 1, not {value}"`?

pybaseball/analysis/trajectories/batted_balls/calculator.py
```diff
@@ -0,0 +1,245 @@
...
+        :param spin_angle: float
+        :param launch_angle: float
+        :param launch_direction_angle: float
+        :return: numpy array
+        """
+        # trajectory_vars = x, y, z, vx, vy, vz
+        _, _, _, vx, vy, vz = trajectory_vars
+        v = np.sqrt(vx ** 2 + vy ** 2 + vz ** 2)
+
+        sidespin = spin * sin_in_degrees(spin_angle)
```
schorrm, line 195: Would it make sense to just convert spin_angle and launch_angle to radians once and then streamline the rest of these calls?

pybaseball/analysis/trajectories/batted_balls/parameters.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+
+@attr.s(frozen=True, kw_only=True)
+class UnitConversions:
+    # conversions,
+    mph_to_fts = attr.ib(default=1.467)
+    ft_to_m = attr.ib(default=0.3048037)
+    lbft3_to_kgm3 = attr.ib(default=16.01848)
+    kgm3_to_lbft3 = attr.ib(default=0.06242789)
```
schorrm, line 73: Can we drop the `attr.ib` in favor of annotating these as float?
https://www.attrs.org/en/latest/types.html

pybaseball/analysis/trajectories/batted_balls/parameters.py
```diff
@@ -0,0 +1,73 @@
...
+    vwind = attr.ib(default=0, metadata={"units": "mph"})  # mph
+    phiwind = attr.ib(default=0, metadata={"units": "deg"})  # deg
+    hwind = attr.ib(default=0, metadata={"units": "ft"})  # ft
+    relative_humidity = attr.ib(default=50)
+    pressure_in_hg = attr.ib(default=29.92)
+    temperature_f = attr.ib(default=70, metadata={"units": "F"})  # F
+    elevation_ft = attr.ib(default=15, metadata={"units": "ft"})
+    beta = attr.ib(
+        default=1.217e-4, validator=check_greater_zero, metadata={"units": "per_meter"}
+    )
```
schorrm, line 43: Do we use the `{"units": ... }` anywhere?

pybaseball/analysis/trajectories/batted_balls/parameters.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.air_density = 1.2929 * (
+            273
+            / (self.temperature_c + 273)
+            * (
+                self.pressure_mm_hg * np.exp(-self.beta * self.elevation_m)
+                - 0.3783 * self.relative_humidity * self.SVP * 0.01
+            )
+            / 760
+        )
```
schorrm, line 64: 0K = 273? Not 273.15?
And then why is the the whole thing here done to in mmhg and then Torr?
Can we streamline the unit system?

pybaseball/analysis/trajectories/batted_balls/parameters.py
TheCleric: Also since these are being treated like constants I'd argue they should be in all caps like MPH_TO_FTS.

https://www.python.org/dev/peps/pep-0008/#constants

pybaseball/analysis/trajectories/batted_balls/parameters.py
bdilday: no, they effectively play the role of comments / documentation

pybaseball/datahelpers/postprocessing.py
bdilday: done

pybaseball/analysis/trajectories/batted_balls/calculator.py
bdilday: What I did here was add `lru_cache` decorators to `sin_in_degrees` and `cos_in_degrees`. And internally they call a cached conversion from degrees to radians. This way the conversion only happens once, which I think captures the optimization you're referring to in this comment?

pybaseball/analysis/trajectories/batted_balls/parameters.py
bdilday: so if you look at the implementation of this in the `pybbda` package, you can see I had the class as a `Singleton`, https://github.com/bdilday/pybbda/blob/master/pybbda/analysis/trajectories/batted_balls/parameters.py#L69 so it was effectively a immutable data-class. If it's not a singleton, then you're right it doesn't make sense to store constants this way. We can get sort of the same immutable behaviour by making it a module, so what I did here was move them to `pybaseball.analysis.trajectories.unit_conversions` and use the upper case convention

bdilday: regarding using type annotations instead of `attr.ib`, this seems like a change from explicit to implicit without any benefit. I'm not really enthusiastic about making that change. Note that the attr documentation for types says, 

> In practice though, types show their biggest usefulness in combination with tools like mypy or pytype that both have dedicated support for attrs classes.

i.e. we already get the benefit of mypy by using `attr.ib`

pybaseball/analysis/trajectories/batted_balls/parameters.py
bdilday: it's done this way because I transcribed it from Alan Nathan's [excel-based calculator](http://baseball.physics.illinois.edu/trajectory-calculator-new3D.html), and that;s how it is there.  I think changing from `273` to `273.15` is more correct and we should go ahead and change it. As far as units, if we change units here, we'll have to change the numerical constants in other parts of the code, which I'm not inclined to do. I think more important than having this be as concise as possible is that it matches a reference, which in this case is Alan Nathan's code. Maybe I'm misunderstanding the comment about streamlining?

pybaseball/analysis/trajectories/batted_balls/calculator.py
schorrm: I'm less bothered about the runtime cost. It's more that it'd be a lot cleaner I think to shuttle everything into radians once and then just have `sin`, `cos`, etc.

bdilday: if you look at where the `sin_in_degrees` and `cos_in_degrees` were being used it was basically two places - making a velocity vector from the launch angles, and decomposing the spin into x, y, z components. To address this comment, I refactored these both into `utils`. so now the trajectpry calculator doesn't mess with angles at all, it just calls `unit_vector` and `spin_components`, and inside of those I took your advice and converted to radians once, and then called the numpy `sin` and `cos`.

pybaseball/analysis/trajectories/batted_balls/parameters.py
bdilday: updated the celsius-to-kelvin conversion

pybaseball/analysis/trajectories/utils.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+from functools import lru_cache
+from pybaseball.analysis.trajectories.unit_conversions import (
+    REVS_PER_MINUTE_TO_RADIANS_PER_SECOND,
+)
+
+
+@lru_cache(maxsize=64)
```
schorrm, line 8: Do we still need the LRU caches?

pybaseball/analysis/trajectories/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    returns the x,y, z components of spin
+
+    :param spin:
+    :param spin_angle:
+    :return:
+    """
+
+    spin_angle_in_radians = np.deg2rad(spin_angle)
+    launch_angle_in_radians = np.deg2rad(launch_angle)
+    launch_direction_angle_in_radians = np.deg2rad(launch_direction_angle)
```
schorrm, line 23: can we remove the `_in_radians`?
Just `spin_angle = np.deg2rad(spin_angle)` etc?
It'll make it cleaner I think

pybaseball/analysis/trajectories/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    wx = (
+        backspin * np.cos(launch_direction_angle_in_radians)
+        - sidespin * np.sin(launch_angle_in_radians) * np.sin(launch_direction_angle_in_radians)
+    ) * REVS_PER_MINUTE_TO_RADIANS_PER_SECOND
+    wy = (
+        - backspin * np.sin(launch_direction_angle_in_radians)
+        - sidespin * np.sin(launch_angle_in_radians) * np.cos(launch_direction_angle_in_radians)
+    ) * REVS_PER_MINUTE_TO_RADIANS_PER_SECOND
+    wz = sidespin * np.cos(launch_angle_in_radians) * REVS_PER_MINUTE_TO_RADIANS_PER_SECOND
```
schorrm, line 36: Can the RPM to Rad/sec conversion be applied just to the `spin` parameter once, and then not apply it elsewhere?
And what do you think about changing the name to `RPM_TO_RAD_SEC`?

pybaseball/analysis/trajectories/utils.py
bdilday: we need it for `spin_components`, because we expect it to be called with the same values of `spin`, `spin_angle`, `launch_angle` and `launch_direction_angle`. i.e. wx, wy, wz are based on the initial spin and angles, and don't change throughout the computation of the trajectory. as with the celsius to kelvin conversion, this is because it's how it's done in Alan Nathan's code and I just transcribed it.

we don't need it for `initial_velocity`

pybaseball/analysis/trajectories/utils.py
bdilday: hmm, I don't like this for two reasons

1. I prefer not to overwrite existing variables
2. I prefer explicit variable names

I'd argue since this is a util function that's not really public facing, and it's functional, let's just leave it as is.

pybaseball/analysis/trajectories/utils.py
bdilday: yes, good call. done.

pybaseball/analysis/trajectories/utils.py
schorrm: Fair enough

schorrm:
I made one tiny tweak to formatting, once CI passes I'll merge

Issue #105, opened 2020-09-04
Add date support to prostproccess.try_parse


TheCleric:
Converting dates would be needed in this function to support dates in data such as with Statcast.

I was reminded of this by #104

TheCleric:
Fixed via #106

Issue #104, opened 2020-09-04
Statcast Game Date Format Problem


jldbc:
Did the statcast csv change its date format? Someone tagged me in this on Twitter https://twitter.com/ckurcon/status/1301913190465507328

I get: `ValueError: time data "2020-09-03T00:00:00.000Z" doesn't match format specified`

TheCleric:
Doesn't appear to be the case for me. I have a statcast file from 9/1 in my saves that shows date like "2020-09-01", and I just ran a couple files with past and current/future dates and see the same format.

TheCleric:
This is my copy of the repo, but is synced with master (which is admittedly slightly out of date with the released version, but nothing around Statcast or dates to my knowledge):

```
Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybaseball
>>> pybaseball.statcast('2020-08-25', '2020-09-03').head()
This is a large query, it may take a moment to complete
Completed sub-query from 2020-08-25 to 2020-08-30
Completed sub-query from 2020-08-31 to 2020-09-03
  pitch_type  game_date  ...  if_fielding_alignment  of_fielding_alignment
0         SL 2020-09-03  ...               Standard               Standard
1         SL 2020-09-03  ...               Standard               Standard
2         SI 2020-09-03  ...               Standard               Standard
3         FS 2020-09-03  ...          Infield shift              Strategic
4         SI 2020-09-03  ...          Infield shift              Strategic

[5 rows x 89 columns]
```

jldbc:
Maybe they fixed it? I got the error an hour ago but don't get it anymore

TheCleric:
Strange. In the mentioned ticket above we can try to support both date formats in the new try_parse parsing function and cut Statcast parsing over to that function (which would also fix #59 as well).

Maybe I'll tackle that next.

jldbc:
Looks like this was a one-off thing, I'm going to close this. 

Some [additional context](https://twitter.com/darenw/status/1301987012191367168) from darenw (doesn't explain why this was happening, but looks like it was unintended behavior that shouldn't be showing up again)

PR #106, opened 2020-09-05
Converts statcast over to using the try_parse code that autodetermines data types

TheCleric:
This PR was done in three steps:

1. Create unit tests for existing statcast code
2. Migrate statcast to using the try_parse postprocessor to autodetermine data types
3. Ensure unit tests run successfully

When merged this PR should fix the following issues: #105, #104, #59


pybaseball/datahelpers/postprocessing.py
```diff
@@ -1,30 +1,64 @@
...
 import numpy as np
 import pandas as pd
 
 null_regexes = [r'^\s*$', r'^null$']
+date_formats = [
+    '%Y-%m-%d',              # Standard statcast format
+    '%Y-%m-%dT%H:%M:%S.%fZ', # Just in case (https://github.com/jldbc/pybaseball/issues/104)
+]
+
+def try_parse_dataframe(
```
schorrm, line 14: Can we make this fit on one line? Maybe by setting something for `Unon[str, int, float', datetime']`?

schorrm: Define `ColType = Union[str, int, float', datetime']` somewhere, and annotate ColType?

TheCleric: Do you find it less readable on multiple lines? I'll put it on one line if desired, but I find it more readable this way. You near instantly can see how many parameters it takes, of what type, and which have defaults. Whereas one one line they tend to somewhat run together in my opinion.

Thoughts?

TheCleric: Also digging through the PEP to see what are the preferred way to handle this and found this discussion:

https://github.com/python/typing/issues/433

Where they suggested an alternative that would look like this if you'd prefer (though I personally find it less readable):

```python
def try_parse_dataframe(data: pd.DataFrame, null_replacement: Union[str, int, float, datetime] = np.nan,
                        known_percentages: List[str] = []) -> pd.DataFrame:
```

schorrm:
LGTM

Issue #87, opened 2020-08-30
Feature: Local caching of data


TheCleric:
This library relies on a set of data sources that are all online, which can be problematic for a few reasons:

1. You may wish to use it offline 

and/or

2. Some of the data source it supports have rate limits (understandably so) that can result in data to not be returned when hit too frequently.

So after discussing with @schorrm in #85 we'd like to add a caching layer that would allow for data to be stored locally for repeat calls. I have some ideas that I will list in further comments, but other suggestions are welcome as well.

schorrm:
So my usual use pattern is one time use of the acquisition functions followed by saving as a CSV / TSV. That's arguably kinda clunky, it making it all 'just work' is arguably the better way.

TheCleric:
One approach would be something along what https://github.com/bdilday/pybbda does. From a cursory analysis of their code base, their approach seems to be, get the data initially, and then reuse that data.

Positives:
- Ensure offline use
Negatives:
- Not as flexible. Rigid process of downloading data before use.

schorrm:
So now on to the meat?
How do we interface with this? Do we have separate functions for downloading and loading from cache, or do we have a unitary function which checks to see if it can pull it from the cache and then executes the query if it can't find it?
We now face the following questions.
1. Where does this go? I suggest `$HOME/.pybaseball/` (we'll need to be careful about the platform independence, e.g. `os.path.join`, not `f'{}/{}'`).
2. How do we track the items in there? We need to standardize a reference form of the query. My initial inclination is t
3. How do we mark an item as invalid in the cache? What are the criteria for an invalid item?
4. Let's say I executed a query or downloaded the data for batters, 2010-2018, independent, no min PA.
I then query batters 2012-2016, independent, min 300 PA. That can be done entirely within the the first file, do we want to make sure that we ID that and get that?

TheCleric:
Another approach is a more passive/silent approach. For example one thing I've done locally is a function wrapper that will cache the data:

```python
def _get_f_hash(f, args, kwargs, by_name_only=False):
    f_hash = f"{f.__name__}" + '' if by_name_only else f"<{f.__hash__()}>"
    
    args_hash = ', '.join([f"{x if isinstance(x, str) else x.__hash__()}" for x in args])
    kwargs_hash = ', '.join(
        [f"{key}={kwargs[key] if isinstance(kwargs[key], str) else kwargs[key].__hash__()}" for key in kwargs.keys()]
    )

[...]

PR #114, opened 2020-09-10
Objectify FanGraphs

TheCleric:
This PR converts all the FanGraphs calls under a unified class.  This accomplishes a few things:

1. Splits out the HTML parsing so that it can be extended for other data sources such as BRef now or Seamheads later.
2. Since all the FanGraphs data revolves around the `/leaders.aspx` page with different params, this collapses them into one root function.
3. Specifies a lot of the options via enums so that type checking is a bit easier. Included in this are complete enums of the supported FanGraphs columns for Batting, Fielding, and Pitching.

In addition I fixed a bug I found in the Lahman reader on Windows.


pybaseball/batting_leaders.py
```diff
@@ -1,51 +1,34 @@
...
+    end_season   : int           : final season you want data for
+    league       : str           : "all", "nl", or "al"
+    qual         : Optional[int] : minimum number of plate appearances to be included in the data.
+                                   If None is specified, the FanGraphs default ('y') is used.
+                                   Default = None
+    ind          : int           : 1 if you want individual season-level data
+                                   0 if you want a player's aggreagate data over all seasons in the query
+    """
+
+    warnings.warn("batting_stats is deprecated in favor of FanGraphs().batting_stats", DeprecationWarning)
```
schorrm, line 26: I do not like the deprecating API change

pybaseball/batting_leaders.py
```diff
@@ -1,51 +1,34 @@
...
+
+    return FanGraphs().batting_stats(
```
schorrm, line 28: It takes, let's say, 10-20 seconds to scrape the thingy, can't we just the 10 microseconds of class constructor overhead?

pybaseball/enums/fangraphs/month.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum, unique
+
+
+@unique
+class FanGraphsMonth(Enum):
+    ALL               = 0
+    MARCH_APRIL       = 4
```
schorrm, line 7: ah, yes, my birth month, the month of March-April.
(I'm aware it's like this there, I'm just laughing at it)

pybaseball/lahman.py
```diff
@@ -33,109 +38,109 @@ def download_lahman():
 def parks():
 	# do this for every table in the lahman db so they can exist as separate functions
 	z = get_lahman_zip()
-	f = os.path.join(base_string, "Parks.csv")
+	f = get_base_string(_separator) + _separator + "Parks.csv"
```
schorrm, line 41: I have a _feeling_ like this should be doable more cleanly, can you show me your original bug? I might have an idea for this.

pybaseball/team_batting.py
```diff
@@ -17,18 +18,15 @@ def team_batting(start_season: int, end_season: int = None, league: str = 'all',
...
-        raise ValueError(
-            "You need to provide at least one season to collect data for. Try team_batting(season) or team_batting(start_season, end_season)."
-        )
-    if end_season is None:
-        end_season = start_season
 
-    fg_data = fangraphs.get_fangraphs_tabular_data_from_url(
-        _FG_TEAM_BATTING_URL.format(start_season=start_season, end_season=end_season, league=league, ind=ind)
-    )
+    warnings.warn("team_batting is deprecated in favor of FanGraphs().team_batting", DeprecationWarning)
```
schorrm, line 22: Same. I'm not sure about deprecating the API here. It's nice if it cleans up the back-end implementation, but I don't love the new API here

pybaseball/team_fielding.py
```diff
@@ -20,20 +19,15 @@ def team_fielding(start_season: int, end_season: int = None, league: str = 'all'
...
     """
 
-    if start_season is None:
-        raise ValueError(
-            "You need to provide at least one season to collect data for. " +
-            "Try team_fielding(season) or team_fielding(start_season, end_season)."
-        )
-    if end_season is None:
-        end_season = start_season
+    warnings.warn("team_fielding is deprecated in favor of FanGraphs().team_fielding", DeprecationWarning)
```
schorrm, line 22: ditto

schorrm:
Given how much is being changed here, I would like feedback from other core contributors. @jldbc , @bdilday , @andersonfrailey -- want to weigh in?

pybaseball/lahman.py
TheCleric: I hadn't experienced it before because I had downloaded the lahman files, but I recently deleted them and now running the marcel unit tests fail on my Windows machine, all with errors similar to this one:

```
tests\pybaseball\analysis\projections\marcels\test_marcels_pitching.py:17:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pybaseball\analysis\projections\marcels\marcels_base.py:35: in __init__
    self.stats_df = stats_df if stats_df is not None else self._load_data()
pybaseball\analysis\projections\marcels\marcels_pitching.py:19: in _load_data
    return pitching()
pybaseball\lahman.py:155: in pitching
    data = pd.read_csv(f if z is None else z.open(f), header=0, sep=',', quotechar="'")
C:\Python38\lib\zipfile.py:1514: in open
    zinfo = self.getinfo(name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  

self = <zipfile.ZipFile file=<_io.BytesIO object at 0x000001B95CE3F9F0> mode='r'>
name = 'baseballdatabank-master\\core\\Pitching.csv'

    def getinfo(self, name):
        """Return the instance of ZipInfo given 'name'."""
        info = self.NameToInfo.get(name)
        if info is None:
>           raise KeyError(
                'There is no item named %r in the archive' % name)
E           KeyError: "There is no item named 'baseballdatabank-master\\\\core\\\\Pitching.csv' in the archive"

C:\Python38\lib\zipfile.py:1441: KeyError
```

pybaseball/batting_leaders.py
TheCleric: We can certainly get rid of it if this isn't desired. It just seemed odd to me to consolidate the FanGraphs code (which I thought we had discussed doing), but then have to continue maintaining these separate implementations. And to be clear, they still work, you just get a warning. I'm not planning on deleting them anytime soon, just trying to get users over to the new methods here.

pybaseball/lahman.py
TheCleric: Now that I think about it, I have a more clean solution in my branch where I've been working on caching. I can bring that one in if you'd like.

TheCleric:
Actually now that I look at this PR with the fresh eyes of the morning there are some things I want to add:

1) Update the documentation
2) Add some integration tests
3) Run some profiling against current master vs. this to check performance.

I'll get those added soon, but only if we think this change is worth pursuing, so I'll wait until we get feedback from everyone else.

pybaseball/batting_leaders.py
schorrm: Why though? What's wrong with the old API?

TheCleric: It's not as flexible as this new API, and as well, consolidating all of these together under one roof will make maintaining them easier (adding new functionality, fixing bugs, adapting to FG changes).

pybaseball/lahman.py
bdilday: actually could the lahman bug fix be moved to a separate PR? I know it's a minor point but I feel like it'd help in focusing discussion on the more substantive changes here

TheCleric: Sure, gimme just a few.

TheCleric: It's in #116

TheCleric: Merged into this branch as well.

schorrm:
Can you give a concrete example of an improvement in the new API, and why we shouldn't just it to backend the existing API?

TheCleric:
> 3. Run some profiling against current master vs. this to check performance.

I've made a profiling script that I'll check in via another PR, but here are the results (10 iterations of each FanGraphs function against local data):

```
$ git checkout master && python tests/pybaseball/profiling.py | head -n1 && git checkout feature/objectify_fangraphs && python tests/pybaseball/profiling.py | head -n1
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
         165221893 function calls (165214594 primitive calls) in 155.581 seconds
Switched to branch 'feature/objectify_fangraphs'
Your branch is up to date with 'origin/feature/objectify_fangraphs'.
         165324573 function calls (165317274 primitive calls) in 157.148 seconds
```

So it does appear to be slightly (1%) slower, but I have some ideas on how to get this faster now that I can see where the slow parts are in the profiler.

TheCleric:
> Can you give a concrete example of an improvement in the new API, and why we shouldn't just it to backend the existing API?

Sure!

So current batting_stats takes only four parameters:
- start_season
- end_season
- plate_appearances (qual)
- league
- individual vs aggregate data (ind)

The new function allows all of those, but as well allows filtering by:
- month
- on_active_roster
- minimum_age
- maximum_age
- position

As well in a future PR we could also add any other filtering that FanGraphs supports on their URL such as:
- start_date
_ end_date

schorrm:
Can we just throw those in a kwargs?

schorrm:
And don't sweat a 1% perf regression

TheCleric:
> Can we just throw those in a kwargs?

We could, but at that point we'd just be passing them blindly down, which means no Intellisense/IDE support for parameter help.

It makes more sense to me to not modify the existing function signature, and implement them in the class function as has been done here. That's just my opinion though.

schorrm:
I hear that argument, I would like to have others on my side before I make a breaking API change.

bdilday:
It seems to me this PR is trying to do too much and would be better broken up into several smaller PRs. My understanding is it's trying to solve 5 problems, 

* make the html table parsing more general

* reduce repetition in the fangraphs scrapers since they're all variations on parsing `/leaders.aspx`

* provide enums to help make sense of the fangraphs integer columns in the query params and to simplify iterating over all of them

* expand the type of queries available to fangraphs, e.g. min and max age, month, on_active_roster, etc.

* fix a bug in parsing lahman data on windows

does that seem accurate, @TheCleric? would it be possible / helpful to make each of those it's own PR?

TheCleric:
> I hear that argument, I would like to have others on my side before I make a breaking API change.

I agree completely. Perhaps it's a lack of communication on my part, but I do not consider this a breaking API change. If existing/new code calls any of those functions, they will still receive data:

```
$ python
Python 3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybaseball
>>> pybaseball.team_batting(2019)
    Season          Team  Season 2  Age     G    AB    PA     H  ...    EV    LA  Barrels  Barrel%  maxEV  HardHit  HardHit%  Events
0     2019        Astros      2019   28  2309  5613  6394  1538  ...  88.4  13.3      282    0.062  117.9     1579     0.350    4517
1     2019         Twins      2019   27  2320  5732  6392  1547  ...  89.7  14.9      416    0.093  117.0     1762     0.396    4452
2     2019       Yankees      2019   28  2312  5583  6245  1493  ...  90.0  13.5      371    0.089  120.6     1693     0.404    4190
3     2019       Dodgers      2019   28  2413  5493  6281  1414  ...  89.3  14.5      319    0.075  114.3     1610     0.380    4237
4     2019       Red Sox      2019   29  2455  5770  6475  1554  ...  89.9  12.7      326    0.073  115.0     1841     0.413    4454
5     2019     Nationals      2019   29  2327  5512  6267  1460  ...  88.6  14.1      294    0.069  113.5     1579     0.368    4294
6     2019        Braves      2019   28  2411  5560  6302  1432  ...  89.2  13.2      341    0.082  115.9     1602     0.386    4156
7     2019          Cubs      2019   29  2429  5461  6195  1378  ...  88.9  10.9      311    0.076  117.6     1450     0.356    4071
8     2019       Rockies      2019   28  2452  5660  6288  1502  ...  88.6  12.4      254    0.060  114.1     1487     0.350    4253
9     2019     Athletics      2019   29  2363  5561  6269  1384  ...  89.4  14.2      332    0.078  114.4     1642     0.385    4266
10    2019          Mets      2019   29  2445  5624  6290  1445  ...  89.1  11.1      282    0.066  118.3     1625     0.378    4295
11    2019       Brewers      2019   28  2514  5542  6309  1366  ...  89.7  12.0      310    0.077  117.9     1589     0.393    4045
12    2019          Rays      2019   27  2459  5628  6285  1427  ...  89.5  11.2      308    0.074  116.2     1628     0.390    4177
13    2019  Diamondbacks      2019   28  2418  5633  6315  1419  ...  88.0  11.8      272    0.063  116.3     1495     0.344    4345
14    2019       Indians      2019   28  2321  5425  6124  1354  ...  89.5  12.9      279    0.067  114.4     1563     0.374    4179
15    2019       Rangers      2019   27  2273  5540  6204  1374  ...  89.4  12.5      283    0.070  115.4     1552     0.386    4025
16    2019        Angels      2019   28  2425  5542  6251  1368  ...  88.0  12.9      275    0.064  116.6     1465     0.339    4322
17    2019      Phillies      2019   28  2431  5571  6261  1369  ...  88.8  13.5      271    0.065  116.4     1521     0.363    4189
18    2019     Cardinals      2019   27  2413  5449  6167  1336  ...  87.9  13.6      269    0.066  115.3     1395     0.339    4110
19    2019      Mariners      2019   28  2347  5500  6199  1305  ...  87.4  14.3      269    0.068  113.1     1279     0.322    3972
20    2019       Pirates      2019   27  2441  5657  6228  1497  ...  88.3  10.1      230    0.051  116.2     1526     0.337    4527
21    2019          Reds      2019   28  2462  5450  6100  1328  ...  87.4  13.6      224    0.055  118.3     1305     0.320    4083
22    2019     Blue Jays      2019   27  2354  5493  6091  1299  ...  88.7  14.2      289    0.072  118.9     1472     0.366    4023
23    2019     White Sox      2019   28  2311  5529  6042  1443  ...  89.1  10.8      268    0.066  117.9     1512     0.373    4049
24    2019       Orioles      2019   27  2386  5596  6189  1379  ...  88.0  11.8      248    0.059  114.8     1418     0.336    4221
25    2019        Padres      2019   26  2413  5391  6019  1281  ...  88.9  10.7      261    0.067  115.9     1439     0.371    3879
26    2019        Royals      2019   27  2269  5496  6080  1356  ...  88.0  12.2      241    0.058  115.7     1376     0.331    4160
27    2019        Giants      2019   29  2493  5579  6170  1332  ...  87.8  13.2      242    0.058  114.7     1501     0.357    4210
28    2019        Tigers      2019   28  2316  5549  6039  1333  ...  88.1  14.1      236    0.059  114.2     1463     0.365    4005
29    2019       Marlins      2019   28  2402  5512  6045  1326  ...  88.3   9.1      215    0.052  114.7     1440     0.351    4108

[30 rows x 313 columns]
>>> 
```

There's just a warning in the background that these are deprecated. It's just a warning pointing them to the new way of doing it, not an exception preventing them from doing so.

TheCleric:
> It seems to me this PR is trying to do too much and would be better broken up into several smaller PRs. My understanding is it's trying to solve 5 problems,
> 
> * make the html table parsing more general
> * reduce repetition in the fangraphs scrapers since they're all variations on parsing `/leaders.aspx`
> * provide enums to help make sense of the fangraphs integer columns in the query params and to simplify iterating over all of them
> * expand the type of queries available to fangraphs, e.g. min and max age, month, on_active_roster, etc.
> * fix a bug in parsing lahman data on windows
> 
> does that seem accurate, @TheCleric? would it be possible / helpful to make each of those it's own PR?

I think this is a fair summary. As far as splitting goes, I have already split out number 5 in pr #116 (it's just in this PR because it uses the same code, once that one is merged it will no longer be a diff between this and master).

I could also see doing a quick PR to address 1 (HTML parsing) individually.

The other three for me are much more difficult to separate. The `/leaders` consolidation necessitates shared understanding across the functions, which is where the enums come in. I could probably separate out the new queries, but to me we're essentially getting them for free with the consolidation as they are query params that are typically passed, we just usually pass them empty. This just gives a facility for giving it a value.

schorrm:
Looking this over, a question: why do need the class methods? When I see the `team_batting` or whatever, it doesn't make use of `self` other than `self._leaders()`. What's the benefit of the class over just a utility method?

schorrm:
Also, I merged your profiling PR -- so can you please drop the `.gitignore` from here and bring the branch up to date to cut down on the sheer number of files involved here?

TheCleric:
> Looking this over, a question: why do need the class methods? When I see the `team_batting` or whatever, it doesn't make use of `self` other than `self._leaders()`. What's the benefit of the class over just a utility method?

For me it's really about the class inheritance from the HTMLTable class that will eventually power other classes that rely on HTML table parsing -> DataFrames.

Could this be done in a purely functional way? Sure. How you feel about that (in my experience) is usually indicative of how you feel about object oriented programming in general. The folks that don't like OOP tend to go with the function library route. The people who like OOP tend to go class based. I'll admit that I fall in the latter camp (though not enough to go full Java and make everything a class).

In my OPINION, it boils down to this: if you have a set of functions that all rely on the same base functions, and share data among them (in this case things like root_url and the like), then you're essentially recreating inheritance in a very roundabout way. If it looks like inheritance, it may as well be inheritance.

TheCleric:
@schorrm would it help if I split out the HTMLTable class and Enums to a separate PR, since those are somewhat standalone (though not super useful by themselves) and it would make this PR smaller after that goes through?

schorrm:
Nah, just drop the extraneous stuff. The reason to prefer the function layout is continuity (extending the signature). Shabbat soon, so I'll have a look again Sat nite.

bdilday:
> Looking this over, a question: why do need the class methods? When I see the `team_batting` or whatever, it doesn't make use of `self` other than `self._leaders()`. What's the benefit of the class over just a utility method?

yeah, this is the same question I've been having.

bdilday:
> > Looking this over, a question: why do need the class methods? When I see the `team_batting` or whatever, it doesn't make use of `self` other than `self._leaders()`. What's the benefit of the class over just a utility method?
> 
> yeah, this is the same question I've been having.

I'm not against OOP, BTW, I just don't understand the reasoning behind setting it in this particular way. this is why I think this sort of thing is good as an issue also. from my perspective this code just came out of left field (pun intended) and it'd be helpful to have context about what problem you're trying to solve, why this architecture makes sense, what (if any) alternatives you considered, how it helps us expand to other html sources in the future, etc,

TheCleric:
> > Looking this over, a question: why do need the class methods? When I see the `team_batting` or whatever, it doesn't make use of `self` other than `self._leaders()`. What's the benefit of the class over just a utility method?
> 
> yeah, this is the same question I've been having.

In this case it allows doing a lot of the shared data (root_url, xpaths, etc.) in a constructor, and not have to pass that into the _leaders function every time it's called, and then passed down to the base html functions when they are called. So in essence, I'm trying to keep shared function params at the class level instead of being passed around functionally.

However, I think we've debated this long enough. If you folks don't see the value in doing it this way, I'll change it to a more procedural approach. Just let me know. This isn't the hill I'm willing to die on. 😆

TheCleric:
> I'm not against OOP, BTW, I just don't understand the reasoning behind setting it in this particular way.

Completely fair. I think my last comment was a little too defensive. I've dealt with too many python devs who were in the camp that classes should NEVER be used, so it gets my hackles up sometimes. So I apologize if this came off as attacking anybody. 😄 

> this is why I think this sort of thing is good as an issue also. from my perspective this code just came out of left field (pun intended) and it'd be helpful to have context about what problem you're trying to solve, why this architecture makes sense, what (if any) alternatives you considered, how it helps us expand to other html sources in the future, etc,

It may be a BIT out of left field as I didn't say I was about to do it, but in the caching issue, we did have a discussion about heading towards a more OOP route:

https://github.com/jldbc/pybaseball/issues/87#issuecomment-683426702

And it seemed there everyone seemed to think this would probably be the right direction for the library. But perhaps I misunderstood, or misgauged the actual desire to do this.

So to be clear: I went this route because it seemed to me that I was writing code that shared data, and would have similar libraries that were based upon the same, more general code. To me that indicates an OOP paradigm. It could be done as shared modules and functions, but to me this way made more sense. I think when it boils down to it, for me, this code makes more sense than doing it as modules and functions, but that's probably just my preference/how my brain works. As stated above I can reorient it if we don't think this is the right approach.

TheCleric:
> Nah, just drop the extraneous stuff. The reason to prefer the function layout is continuity (extending the signature). Shabbat soon, so I'll have a look again Sat nite.

Shabbat shalom! 😃

TheCleric:
Also for the record as part of my research for this html processor I also tried out the read_html function in pandas:

https://pandas.pydata.org/docs/reference/api/pandas.read_html.html

I was able to get some passable results with it (would still take some postprocessing around percentages), but found it to be about 2x as slow. On this branch, the profiler took over 100 seconds to use read_html, vs. @ 64 seconds for our custom html table -> dataframe parser.

bdilday:
> However, I think we've debated this long enough. If you folks don't see the value in doing it this way, I'll change it to a more procedural approach. Just let me know. This isn't the hill I'm willing to die on. laughing

I actually do see value in it, I'm just not sure that the class hierarchy is done in the way that's most effective. Here's how I see it - we have `FangraphsDataTable` objects (or more generally, `HTMLDataTable`). The attributes they need are, 

* url to fetch data from
* fixed query parameters 
* data schema (i.e. the enums)

methods they need are
public:
* a "fetch data" method 

private:
* data validator
* data postprocessor

the fetch data method needs to take query parameters, combine them with the fixed parameters, and pass them to the html-accessor to get the data, then do postprocessing and validation and return the result as a data frame. 

so I think where the difference comes in with what you did is

* we can have an abstract base class that implements the data fetch, and a concrete class is created by specifying what the url is, what the fixed parameters are, and the data postprocessor and validators, and the data schema. 

* the html-accessor does not need to be the parent class. in fact it feels to me like it should be a low-level utility module. and that it should be stateless - it just takes urls, query parameters, and table specifications and returns data. and it doesn't matter if you have to specify all the parameters in the function call because it's not public facing, its a low-level utility.

I opened a [PR on your branch](https://github.com/TheCleric/pybaseball/pull/11) @TheCleric  to illustrate how I see it working. it's a branch off of `feature/objectify_fangraphs` to keep the diff focused. for expediency I didn't really change the `HTMLTable` class, even though I said above I thought it should be a low-level utility. unfortunately there's some formatting stuff that got in there since I used `black` on my code

TheCleric:
> I opened a [PR on your branch](https://github.com/TheCleric/pybaseball/pull/11) @TheCleric to illustrate how I see it working. it's a branch off of `feature/objectify_fangraphs` to keep the diff focused on the most relevant changes. for expediency I didn't really change the `HTMLTable` class, even though I said above I thought it should be a low-level utility. unfortunately there's some formatting stuff that got in there since I used `black` on my code

@bdilday at first when I saw this solution, I have to admit: I didn't like it. The more I look at it though, the more sense I see in this approach. A few things I'd like to discuss further though:

- Callable classes feel funky (and unintuitive to an end user of the library, or at least they would to me). I think I'd rather see something like:
```python
FangraphsBattingStats().fetch(
    start_season,
    end_season=end_season,
    league=FanGraphsLeague(league),
    qual=qual,
    split_seasons=(ind=='1')
)
```
rather than:
```python
FangraphsBattingStats()(
    start_season,
    end_season=end_season,
    league=FanGraphsLeague(league),
    qual=qual,
    split_seasons=(ind=='1')
)
```
 
- I was hoping to unify all the FanGraphs calls into a single object. This is just a personal preference, and probably part of a larger debate about how functionality should be grouped: by data source, by result data genre, or some other way? I.e., should we have groupings of:
    - All FanGraphs calls unified, all BRef calls unified, etc. (Similar to how Lahman works now).
    - OR All the batting data together (regardless of source), all pitching data together, etc.
    - OR some other organization system

I think grouping like calls makes for a better user experience at it becomes more intuitive on where to find similar calls.

Let me know what you think as well as what @schorrm thinks about these.

bdilday:
@TheCleric  re callable classes and having the calls unified, sorry I didn't finish my thought on that. what I intended was that we would expose the classes with an alias to match the old API, e.g., `batting_stats = FangraphsBattingStats()`, then a user can do

```
from pybaseball.datasources.fangraphs import batting_stats, pitching_stats
df = batting_stats(2009)
```
just as before. that was the motivation for making the class callable and it makes the calls unified in the `pybaseball.datasources.fangraphs` module.

I opened a new PR (in my fork) that shows this and also gets rid of all the noise in the diffs due to formatting differences https://github.com/bdilday/pybaseball/pull/1/

TheCleric:
> @TheCleric re callable classes and having the calls unified, sorry I didn't finish my thought on that. what I intended was that we would expose the classes with an alias to match the old API, e.g., `batting_stats = FangraphsBattingStats()`, then a user can do
> 
> ```
> from pybaseball.datasources.fangraphs import batting_stats, pitching_stats
> df = batting_stats(2009)
> ```

I don't see a real advantage to exposing the functions that way. We're still keeping the old API, so that can be used if desired, but to continue that naming paradigm within the new library seems clunky, and I would argue could lead to bugs. Since they do not have the same function signature, if someone were to migrate it by just changing the import, it would not work. I'm definitely interested in bringing some of your ideas from that PR in (such as the ABC class), but I would much rather not expose the same name functions in the fangraphs module. As I said I would prefer having the existing function of `batting_stats(...)` continue to work (with the deprecation warning), and having something like `FangraphsBattingStats().fetch(...)`, that way there is a clean delineation between the old functionality and this new one.

TheCleric:
Although I am amenable to the argument that there will be no params passed to the constructor (at least at this time, and for the forseeable future), so making the end user initialize it pointlessly just to access the function does seem cumbersome. Is there a middle ground here where the last lines of fangraphs.py go something like:

```python
batting_stats_advanced = FangraphsBattingStats()
pitching_stats_advanced = FangraphsPitchingStats()
team_batting_advanced = FangraphsTeamBattingStats()
team_pitching_advanced = FangraphsTeamPitchingStats()
team_fielding_advanced = FangraphsTeamFieldingStats()
```

Not a big fan of `_advanced` as the delineating word there, but you get my point.

Also I wish there was a better way to combine the utility of class inheritance here without the useless constructor. I'm sure there is a way to simulate it via module methods, but to me that would be even more hacky.

TheCleric:
Also going the route above means singletons should probably be used.

schorrm:
aaaaaagggggghhhhhhhh

schorrm:
this going way too fast to "last hw assignment of the semester"

schorrm:
If we have a class with a method and without much data, if we want the method inheritance, that should be a fairly cheap to do so? the way @bdilday is suggesting?

TheCleric:
> If we have a class with a method and without much data, if we want the method inheritance, that should be a fairly cheap to do so? the way @bdilday is suggesting?

Yeah I think so. If we think that's a good middle ground, I'll implement some of his suggestions. As well, I'm almost tempted to close the PR and start a fresh one. Let me know what you think @schorrm

TheCleric:
After chatting with @schorrm I'm going to close this PR, implement some of @bdilday 's suggestions, and break this into a couple smaller PRs.

Thanks for the feedback folks.

PR #116, opened 2020-09-10
Refactor Lahman and fix separator bug in windows

TheCleric:
There is presently a bug in the lahman library for Windows users with no downloaded cache of the data. This is due to the fact that we use os.sep to build the file paths, but internal to the zip file, it is always `/` regardless of os.

In addition, I've refactored this code to remove a lot of the common code among the functions.


.gitignore
```diff
@@ -145,3 +145,6 @@ cython_debug/
 
 # MacOS
 .DS_Store
+
+# lahman cache
+baseballdatabank-master
```
schorrm, line 150: Whoa whoa whoa

TheCleric: Is this not preferred? I'm just preventing an accidental checkin of anyone downloading lahman while working on the library. If we want to leave this to enforcement at the PR level we certainly can. Just trying to add some security.

schorrm:
I think there's a much cleaner way to go about this, mind if I take a shot?

.gitignore
schorrm: fair enough, but this **really** shouldn't happen

schorrm:
I think it made a bunch of things overly complex

schorrm:
Closing, #118

PR #117, opened 2020-09-10
Add a profiler; improve try_parse performance

TheCleric:
This PR includes a profiler that will help us measure performance gains/losses.

As well, I found a slowdown in using strptime to determine when a value was a date, so I've added regex for date determination that seem to cut the time to process FanGraphs data in half:

Before:
```
$ python tests/pybaseball/profiling.py | head -n1
         165221893 function calls (165214594 primitive calls) in 155.581 seconds
```

After:
```
$ python tests/pybaseball/profiling.py | head -n1
         37690404 function calls (37683256 primitive calls) in 71.389 seconds
```


schorrm:
LGTM.

PR #118, opened 2020-09-11
Alternate refactor for Lahman

schorrm:
Different refactor for Lahman.
Keeping existing utility functions as-is, simplify the reuse, and add type annotations.
Fix the zipfile bug by just using `/` only, yes, it works just fine on Windows.
(Instead of #116)

### DIFF
--- a/pybaseball/lahman.py
+++ b/pybaseball/lahman.py
@@ -1,5 +1,5 @@
 import requests
-import zipfile
+from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


TheCleric:
Looks good to me, and works on my windows machine! 😄

TheCleric:
We can use this one and I'll close #116 if you'd like (and integrate this into #114).

PR #120, opened 2020-09-13
Objectify FanGraphs 2.0

TheCleric:
v2.0 of #114 with a lot of feedback from @schorrm and @bdilday integrated in.

**NOTE:** At @schorrm 's request (to keep this PR smaller) I've removed the unit testing. I do have those coded and working. They will come in #121 after this one is merged.

This PR converts all the FanGraphs calls under a unified methodology. This accomplishes a few things:

Splits out the HTML parsing so that it can be extended for other data sources such as BRef now or Seamheads later.
Since all the FanGraphs data revolves around the /leaders.aspx page with different params, this collapses them into one root function.
Specifies a lot of the options via enums so that type checking is a bit easier. Included in this are complete enums of the supported FanGraphs columns for Batting, Fielding, and Pitching.


schorrm:
No, I didn't mean remove the unit testing, I meant leave it as is (and intentionally fail it, just so the file count wouldn't be so high)

TheCleric:
> No, I didn't mean remove the unit testing, I meant leave it as is (and intentionally fail it, just so the file count wouldn't be so high)

Sorry. I'm allergic to broken unit tests. 🤣 

I'll revert them shortly.

I'm not sure if the PRs will be mergable with broken tests, so let me know if I need disable the tests check as well and re-enable them in the next PR.

schorrm:
It better be mergable with broken unit tests. I want the diagnostics, I don't want the CI telling _me_ what to do. M

TheCleric:
> It better be mergable with broken unit tests. I want the diagnostics, I don't want the CI telling _me_ what to do. M

I do. Most developers I know can't be trusted. 🤣

pybaseball/batting_leaders.py
```diff
@@ -1,51 +1,34 @@
...
+    """
+
+    warnings.warn("batting_stats is deprecated in favor of fg_batting_data", DeprecationWarning)
+
+    return fg_batting_data(
+        start_season,
+        end_season=end_season,
+        league=FanGraphsLeague(league),
+        qual=qual,
+        split_seasons=(ind=='1')
```
schorrm, line 33: okay whoaaa what is this?

schorrm, line 33: if this is the same, why isn't `batting_stats` the same as `fg_batting_data`? why do we need separate hooks for this?

pybaseball/datasources/fangraphs.py
```diff
@@ -1,44 +1,135 @@
...
+        url_options = {
+            'pos': position.value,
+            'stats': self.STATS.value,
+            'lg': league.value,
+            'qual': qual if qual is not None else 'y',
+            'type': types,
+            'season': end_season,
+            'month': month.value,
+            'season1': start_season,
+            'ind': '1' if split_seasons else '0',
```
schorrm, line 76: aka `str(int(split_seasons))` or for that matter `f'{split_seasons:d}'`, but the whole thing seems unnecessary. Why not just leave this as an int?

schorrm, line 76: aka `str(int(split_seasons))` or for that matter `f'{split_seasons:d}'`, but the whole thing seems unnecessary. Why not just leave this as an int?

pybaseball/datasources/html_table_processor.py
```diff
@@ -0,0 +1,86 @@
...
+            html_dom,
+            column_name_mapper=column_name_mapper,
+            known_percentages=known_percentages
+        )
+
+    def get_tabular_data_from_url(self, url: str, column_name_mapper: ColumnListMapperFunction = None,
+                                  known_percentages: List[str] = []) -> pd.DataFrame:
+        response = requests.get(self.root_url + url)
+
+        if response.status_code > 399:
```
schorrm, line 68: Can this be anything other than 200 and still pass?

pybaseball/batting_leaders.py
schorrm, line 33: what is this?

schorrm, line 33: if this is the same, why isn't `batting_stats` the same as `fg_batting_data`? why do we need separate hooks for this?

pybaseball/team_batting.py
```diff
@@ -17,18 +18,15 @@ def team_batting(start_season: int, end_season: int = None, league: str = 'all',
...
-    if end_season is None:
-        end_season = start_season
 
-    fg_data = fangraphs.get_fangraphs_tabular_data_from_url(
-        _FG_TEAM_BATTING_URL.format(start_season=start_season, end_season=end_season, league=league, ind=ind)
-    )
+    warnings.warn("team_batting is deprecated in favor of fg_team_batting_data", DeprecationWarning)
 
-    return fg_data
+    return fg_team_batting_data(
```
schorrm, line 24: same concern applies

pybaseball/lahman.py
```diff
@@ -1,9 +1,11 @@
...
 import os
-import pandas as pd
+import zipfile
 from io import BytesIO
-from typing import Optional
+from typing import IO, Optional, Tuple, Union
+from zipfile import ZipFile
+
+import pandas as pd
+import requests
```
schorrm, line 8: This should be untouched

pybaseball/pitching_leaders.py
```diff
@@ -1,50 +1,34 @@
...
-                              0 if you want a player's aggreagate data over all seasons in the query
+    start_season : int           : first season you want data for
+                                   (or the only season if you do not specify an end_season)
+    end_season   : int           : final season you want data for
+    league       : str           : "all", "nl", or "al"
+    qual         : Optional[int] : minimum number of plate appearances to be included in the data.
+                                   If None is specified, the FanGraphs default ('y') is used.
+                                   Default = None
+    ind          : int           : 1 if you want individual season-level data
+                                   0 if you want a player's aggreagate data over all seasons in the query
```
schorrm, line 23: this could also be bool, if we're going to change the API around here (and that wouldn't break code that used 0/1)

pybaseball/pitching_leaders.py
```diff
@@ -1,50 +1,34 @@
...
-            start_season=start_season,
-            ind=ind,
-            types=_FG_PITCHING_LEADERS_TYPES
-        )
+    return fg_pitching_data(
+        start_season,
+        end_season=end_season,
+        league=FanGraphsLeague(league),
+        qual=qual,
+        split_seasons=(ind=='1')
```
schorrm, line 33: ??????

tests/pybaseball/analysis/projections/marcels/test_marcels_pitching.py
```diff
@@ -1,5 +1,5 @@
-from pybaseball.analysis.projections.marcels import MarcelProjectionsPitching
-from pandas import DataFrame
+from pybaseball.analysis.projections.marcels import MarcelProjectionsPitching	
+from pandas import DataFrame
```
schorrm, line 2: This file should not be in the PR at all, what happened here?

tests/pybaseball/analysis/projections/marcels/test_marcels_batting.py
```diff
@@ -1,5 +1,5 @@
-from pybaseball.analysis.projections.marcels import MarcelProjectionsBatting
-from pandas import DataFrame
+from pybaseball.analysis.projections.marcels import MarcelProjectionsBatting	
+from pandas import DataFrame
```
schorrm, line 2: This file should not be in the PR at all, what happened here?

schorrm: Also you just fudged the whitespace. Why did you add these to the commit in the first place?

pybaseball/team_pitching.py
```diff
@@ -18,19 +18,16 @@ def team_pitching(start_season: int, end_season: int = None, league: str = 'all'
...
-        )
-    if end_season is None:
-        end_season = start_season
-
-    fg_data = fangraphs.get_fangraphs_tabular_data_from_url(
-        _FG_TEAM_PITCHING_URL.format(start_season=start_season, end_season=end_season, league=league, ind=ind)
+    
+    warnings.warn("team_pitching is deprecated in favor of fg_team_pitching_data", DeprecationWarning)
+
+    return fg_team_pitching_data(
```
schorrm, line 24: same

pybaseball/team_pitching.py
```diff
@@ -18,19 +18,16 @@ def team_pitching(start_season: int, end_season: int = None, league: str = 'all'
...
+        start_season,
+        end_season=end_season,
+        league=FanGraphsLeague(league),
+        split_seasons=(ind==1)
```
schorrm, line 28: why?

pybaseball/datasources/html_table_processor.py
```diff
@@ -0,0 +1,86 @@
...
+
+        return self.get_tabular_data_from_element(
+            html_dom,
+            column_name_mapper=column_name_mapper,
+            known_percentages=known_percentages
+        )
+
+    def get_tabular_data_from_url(self, url: str, column_name_mapper: ColumnListMapperFunction = None,
+                                  known_percentages: List[str] = []) -> pd.DataFrame:
+        response = requests.get(self.root_url + url)
```
schorrm, line 66: Is there a reason we don't use the format `requests.get(url_endpoint, params=mydict)`?

schorrm (changes requested):

pybaseball/datasources/fangraphs.py
TheCleric: Because it can only ever be a 1 or 0. So it's really a boolean.

pybaseball/datasources/html_table_processor.py
TheCleric: In theory yes, technically anything in the 2XX range is a success. As well anything in the 3XX range could be a redirect, though requests should follow those automatically to the real page.

So according to HTTP standards only status_codes >= 400 are errors. In actual practice we'll likely only be getting things in the 2XX range, 4XX range, or 5XX range. So we could make this `>299` just to catch weird redirects that requests missed, but otherwise the above is fairly standard.

pybaseball/batting_leaders.py
TheCleric: Me not respecting the data type! 😆 

I'll fix.

TheCleric: Existing code wants an int. This is not quite the correct data type for this field as it's either on (1) or off(0). So we're translating the existing data type to booleans to be more honest and specific about what that function accepts.

Also I need to compare to an int, not a string, which I will fix.

pybaseball/lahman.py
TheCleric: I'll undo. Sorry. Must have organized these imports.

pybaseball/pitching_leaders.py
TheCleric: It's true python would just use the truthiness of the 1 or 0, so the effect SHOULD be the same, for anyone who uses MyPy against their code to ensure data type consistency, that would raise an error.

I was erring on the side of not breaking existing function declarations for end user consistency.

tests/pybaseball/analysis/projections/marcels/test_marcels_pitching.py
TheCleric: Must have been checking something and some whitespace got added. Will fix.

tests/pybaseball/analysis/projections/marcels/test_marcels_batting.py
TheCleric: Because you requested that I re-add the old testing, so these two files got caught up in the mix. Since I was just re-adding old files, I did not evaluate them individually.

pybaseball/datasources/html_table_processor.py
TheCleric: Because I didn't know about that. Will add it in.

pybaseball/datasources/fangraphs.py
schorrm: okay, and to convert to int to pass it in requests just use `int(split_seasons)`, I guess

pybaseball/batting_leaders.py
schorrm: That was addressing the ohter one -- why are separating batting_stats and fg_batting_data?

TheCleric: batting_stats is the existing api. fg_batting_data is the new api. I did not want them named the same as I didn't want to cause confusion since they accept a different set of parameters.

If a user wants to continue using the old API it will continue to work. If they want to use the new one, it will be a conscious decision on their part.

pybaseball/batting_leaders.py
```diff
@@ -30,5 +30,5 @@ def batting_stats(start_season: int, end_season: int = None, league: str = 'all'
         end_season=end_season,
         league=FanGraphsLeague(league),
         qual=qual,
-        split_seasons=(ind=='1')
+        split_seasons=(ind==1)
```
schorrm, line 33: bool(ind)?

pybaseball/datasources/fangraphs.py
```diff
@@ -1,44 +1,135 @@
...
-            postprocessing.try_parse(y, headings[index], known_percentages=known_percentages)
-            for index, y in enumerate(x.xpath('td[position()>1]/descendant-or-self::*/text()'))
-        ]
-        for x in data_rows_dom
-    ]
+    def _validate(self, data: pd.DataFrame) -> pd.DataFrame:
+        return data
 
-    fg_data = pd.DataFrame(data_rows, columns=headings)
+    def fetch(self, start_season: int, end_season: Optional[int] = None, types: FanGraphsStat = None,
```
bdilday, line 48: I know I'm the one who removed it in my example PR, but could we put the doc string back in here?

bdilday, line 48: how does using a non-default value for `stats` work? I tried calling it with `fg_batting_data(2019, types=FanGraphsBattingStat.BABIP.value)` and got a 500 error. I guess you need the `c` to be inserted before the list of integer stats codes?

pybaseball/datasources/fangraphs.py
```diff
@@ -1,44 +1,135 @@
...
+        )
+
+class FanGraphsBattingDataTable(FanGraphsDataTable):
+    STATS: FanGraphsStatTypes = FanGraphsStatTypes.BATTING
+    DEFAULT_TYPES: FanGraphsStat = FanGraphsBattingStat.ALL()
+    COLUMN_NAME_MAPPER: ColumnListMapperFunction = BattingStatsColumnMapper().map_list
+    KNOWN_PERCENTAGES: List[str] = ["GB/FB"]
+
+    def _postprocess(self, data: pd.DataFrame) -> pd.DataFrame:
+        return data.sort_values(["WAR", "OPS"], ascending=False)
```
bdilday, line 104: If a user asks for non-default list of stats then `WAR` and / or `OPS` might not be valid columns. Could we handel that case here to avoid a possible `KeyError` when `sort_values` looks for a non-existing column?

pybaseball/datasources/fangraphs.py
bdilday: sorry, I meant non-default value for the argument, `types`

TheCleric: CURSE YOU @bdilday.

Yeah I'll add it back in. I had intended to anyway.

TheCleric: Yeah, it seems you either have to specify the c (which I think stands for common stats), or it's components (Name, Team, Season). I should at least document that, and possibly fix the call to insert that. What do you think?

pybaseball/datasources/fangraphs.py
TheCleric: Very good point. I'll work something in.

pybaseball/datasources/fangraphs.py
TheCleric: And now that I'm thinking further I need to revamp some of this code. If the request is for a fixed set of columns from the user, this shouldn't be a stat coming in, but rather a list of them. Thanks for helping me see a blind spot there.

bdilday: a suggestion - what if as a first step we don't give the user a choice, you just get all the stats. the user is free to filter the columns of the resulting data frame if they want. in a subsequent PR we could fix it, but it seems potentially complicated and I feel like deferring and merging this useful PR first is better than either blocking this PR or rushing a non-optimal solution

pybaseball/datasources/fangraphs.py
```diff
@@ -1,44 +1,135 @@
...
+
+class FanGraphsPitchingDataTable(FanGraphsDataTable):
+    STATS: FanGraphsStatTypes = FanGraphsStatTypes.PITCHING
```
bdilday, line 107: I'm finding it a little confusing that `STATS` has a data type `FanGraphsStatTypes` and `TYPES` has a data type `FanGraphsStat`. Could we somehow make the names more consistent here? One thought is maybe the argument names for the method don't have to match the names in the fangraphs API? i.e. what the `types` query parameter is specifying is maybe better thought of table columns than "stats" (or "types" which is super ambiguous not tp mention conflict with the python protected keyword `type`)?

TheCleric: Valid point. I'll rework.

pybaseball/datasources/fangraphs.py
schorrm: Yeah, I'll change my mind on this and side with @bdilday here

pybaseball/datasources/fangraphs.py
schorrm: Let's defer to as per @bdilday's other suggestion

pybaseball/datasources/fangraphs.py
TheCleric: I didn't see these until after I had done some tweaks here. I now have it working consistently.

For example:

```python
import pybaseball
from pybaseball.enums.fangraphs import FanGraphsBattingStat

pybaseball.datasources.fangraphs.fg_batting_data(2019, types=[FanGraphsBattingStat.AB])
```

Now returns:

|Season | Name | Team | AB
|-- | -- | -- | --
|2019 | Christian Yelich | Brewers | 489
|2019 | Mike Trout | Angels | 470
|2019 | Alex Bregman | Astros | 554
|2019 | Nelson Cruz | Twins | 454
|2019 | Cody Bellinger | Dodgers | 558
|... | ... | ... | ...
|2019 | Khris Davis | Athletics | 481
|2019 | Yolmer Sanchez | White Sox | 496
|2019 | Brandon Crawford | Giants | 500
|2019 | Mallex Smith | Mariners | 510
|2019 | Orlando Arcia | Brewers | 494

Let me know if this works.

pybaseball/datasources/fangraphs.py
TheCleric: This has already been tweaked, but we can discuss further.

pybaseball/datasources/fangraphs.py
bdilday: LGTM!

pybaseball/datasources/fangraphs.py
TheCleric: Names have been reworked to be less confusing.

pybaseball/datasources/fangraphs.py
schorrm: Sorry, I _really_ hate to necromance this, but.
1. This is **really** good behavior to have.
2. We should allow:
```python
import pybaseball
pybaseball.fg_batting_data(2019, types=['AB'])
```

TheCleric: Already in there that way :)

stat_columns accepts either a str or List[str]

```python
import pybaseball

pybaseball.datasources.fangraphs.fg_batting_data(2019, stat_columns=['AB'])
```

|Season | Name | Team | AB
|-- | -- | -- | --
|2019 | Christian Yelich | Brewers | 489
|2019 | Mike Trout | Angels | 470
|2019 | Alex Bregman | Astros | 554
|2019 | Nelson Cruz | Twins | 454
|2019 | Cody Bellinger | Dodgers | 558
|... | ... | ... | ...
|2019 | Khris Davis | Athletics | 481
|2019 | Yolmer Sanchez | White Sox | 496
|2019 | Brandon Crawford | Giants | 500
|2019 | Mallex Smith | Mariners | 510
|2019 | Orlando Arcia | Brewers | 494

135 rows × 4 columns

PR #121, opened 2020-09-14
Objectify Fangraphs v2.0 w/ Unit Tests

TheCleric:
Ignore this until #120 is merged so that it's less of a diff.


pybaseball/datasources/fangraphs.py
```diff
@@ -42,16 +43,55 @@ def __init__(self):
...
-        types = self.DEFAULT_TYPES if types is None else types
+        """
+        Get leaderboard data from FanGraphs.
+
+        ARGUMENTS:
+        start_season       : int                : First season to return data for
+        end_season         : int                : Last season to return data for
+                                                  Default = start_season
+        types              : str                : The columns of data to return
+        league             : FanGraphsLeague    : League to return data for: ALL, AL, FL, NL
```
schorrm, line 70: I think in these cases the API should be a string (= "AL", "NL" etc (pref not case sensitive) and then we should handle the enum

pybaseball/datasources/fangraphs.py
```diff
@@ -42,16 +43,55 @@ def __init__(self):
...
+                                                  Default = FanGraphsMonth.ALL
+        on_active_roster   : bool               : Only include active roster players.
+                                                  Default = False
+        minimum_age        : int                : Minimum player age.
+                                                  Default = 0
+        maximum_age        : int                : Maximum player age.
+                                                  Default = 100
+        team               : str                : Team to filter data by.
+                                                  Specify "0,ts" to get aggregate team data.
+        position           : FanGraphsPositions : Position to filter data by.
```
schorrm, line 88: Same re string. Easy to do '2B', rather than go figure out where the enum is

pybaseball/datasources/fangraphs.py
TheCleric: Implemented.

pybaseball/datasources/fangraphs.py
TheCleric: Implemented.

schorrm:
Gahhh there's a diff bug where it won't update the diff view to the current master

schorrm:
Can you merge back the master or just open this as a clean - single commit - PR?

schorrm:
I'm just trying to clean up the history a bit

TheCleric:
Closing in favor of the cleaner #124

PR #126, opened 2020-09-17
Improve typing in test_statcast to eliminate MyPy errors

TheCleric:


TheCleric:
Tests are failing until I merge master in after #124 gets merged

schorrm:
You're a bit behind. Merge and bring back up to date.

schorrm:
Same thingy, sorry about the history rewrite.

TheCleric:
This should be good now @schorrm

pybaseball/statcast.py
```diff
@@ -117,19 +118,18 @@ def large_request(start_dt, end_dt, d1, d2, step, verbose, team: str = None) ->
...
                     else:
-                        print("Query unsuccessful for data from {} to {}. Skipping these dates.".format(start_dt, tmp_end))
+                        print("Query unsuccessful for data from {} to {}. Skipping these dates.".format(d, tmp_end.strftime(DATE_FORMAT)))
                     if smaller_data_2.shape[0] > 1:
                         dataframe_list.append(smaller_data_2)
-                        print("Completed sub-query from {} to {}".format(intermediate_end_dt, intermediate_end_dt))
+                        print("Completed sub-query from {} to {}".format(d.strftime(DATE_FORMAT), d.strftime(DATE_FORMAT)))
                     else:
-                        print("Query unsuccessful for data from {} to {}. Skipping these dates.".format(intermediate_end_dt, intermediate_end_dt))
+                        print("Query unsuccessful for data from {} to {}. Skipping these dates.".format(d.strftime(DATE_FORMAT), d.strftime(DATE_FORMAT)))
```
schorrm, line 132: Can we please clean this up a bit with f-strings?

pybaseball/statcast.py
```diff
@@ -10,20 +11,22 @@
...
 
-def sanitize_input(start_dt, end_dt):
+def sanitize_input(start_dt: Optional[str], end_dt: Optional[str]) -> Tuple[datetime.datetime, datetime.datetime]:
     # If no dates are supplied, assume they want yesterday's data
     # send a warning in case they wanted to specify
     if start_dt is None and end_dt is None:
         today = datetime.datetime.today()
-        start_dt = (today - datetime.timedelta(1)).strftime("%Y-%m-%d")
-        end_dt = today.strftime("%Y-%m-%d")
+        start_dt = (today - datetime.timedelta(1)).strftime(DATE_FORMAT)
```
schorrm, line 28: If we use `timedelta(days=1)` then we still have a *date* type, not a datetime type, and then just `str(dt)` gives us what we need, not the strftime

pybaseball/statcast.py
```diff
@@ -10,20 +11,22 @@
...
+        end_dt = today.strftime(DATE_FORMAT)
```
schorrm, line 29: Let's remove all of these in favor of just `str()` (see above)

pybaseball/utils.py
```diff
@@ -67,18 +67,18 @@ def sanitize_input(start_dt, end_dt, player_id):
...
 
 
-def split_request(start_dt, end_dt, player_id, url):
+def split_request(start_dt: str, end_dt: str, player_id: int, url: str) -> pd.DataFrame:
     """
     Splits Statcast queries to avoid request timeouts
     """
     current_dt = datetime.datetime.strptime(start_dt, '%Y-%m-%d')
-    end_dt = datetime.datetime.strptime(end_dt, '%Y-%m-%d')
+    end_dt_datetime = datetime.datetime.strptime(end_dt, '%Y-%m-%d')
```
schorrm, line 75: ???

tests/pybaseball/conftest.py
```diff
@@ -1,11 +1,17 @@
...
 import pandas as pd
 import pytest
 import requests
 from _pytest.monkeypatch import MonkeyPatch
+from typing_extensions import Protocol
+
+_ParseDates = Union[bool, List[int], List[str], List[List], Dict]
+
+class GetDataFrameCallable(Protocol):
+    def __call__(self, filename: str, parse_dates: _ParseDates = False) -> pd.DataFrame: ...
```
schorrm, line 14: I'm kinda confused by this

pybaseball/utils.py
TheCleric: MyPy flags this code as not using a best practice. It reuses an existing variable AND changes its type (from str to datetime). This var name change disambiguates it, so we're clear on which is the string and which is the datetime.

tests/pybaseball/conftest.py
TheCleric: Protocols are a way to define that when we pass a function around (like we do in our tests to pass the functions in to load data), that the function sent over will take the correct types, and return the correct types.

So in this instance we're defining that the type `GetDataFrameCallable` is a function that will take the params defined in `__call__` and the return type defined in `__call__`.

pybaseball/utils.py
schorrm: why datetime and not date?

tests/pybaseball/conftest.py
schorrm: Can you take that comment and maybe note where this is used and make that a comment?

pybaseball/statcast.py
TheCleric: `datetime.datetime.today()` is a datetime type, but you raise a lot of good points here that really point out that these shouldn't be datetimes, since we don't care about the time. They should be dates.

pybaseball/utils.py
TheCleric: We're on the same page. Just fixed all datetimes to dates.

tests/pybaseball/conftest.py
TheCleric: Sure!

pybaseball/statcast.py
```diff
@@ -10,20 +11,26 @@
...
 
-def sanitize_input(start_dt, end_dt):
+def sanitize_input(start_dt: Optional[str], end_dt: Optional[str]) -> Tuple[datetime.date, datetime.date]:
     # If no dates are supplied, assume they want yesterday's data
     # send a warning in case they wanted to specify
     if start_dt is None and end_dt is None:
-        today = datetime.datetime.today()
-        start_dt = (today - datetime.timedelta(1)).strftime("%Y-%m-%d")
-        end_dt = today.strftime("%Y-%m-%d")
+        today = datetime.datetime.today().date()
```
schorrm, line 27: `datetime.date.today()`

TheCleric: Confusingly, that's a datetime, not a date:

```
$ python
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> print(str(datetime.datetime.today()))
2020-09-21 16:18:15.435243
```

schorrm: `datetime.date.today()`

schorrm: I was being precise there

pybaseball/statcast.py
schorrm: `datetime.date.today()` is a date type, not a datetime type. Use that instead

pybaseball/statcast.py
TheCleric: Reading fail on my part! Will fix.

TheCleric: Done.

PR #127, opened 2020-09-20
Integration tests for fg_batting_data

TheCleric:
Going to slowly add some integration tests to the project. Here's the first of more to come.

This covers the fg_batting_data function. Per @schorrm 's request, this only brings back a page of 10 when possible so as to keep these lightweight and faster.

I tried to be expansive in terms of testing all possible parameters.

Also I added code coverage to the unit test runner so we'll have a good idea of where we need to shore up our tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,8 +80,8 @@ setup(
                       'requests>=2.18.1',
                       'lxml>=4.2.1',
                       'pygithub>=1.51',
-                      'altair',
-                      'scipy>=1.4.0'
+                      'altair>=4.1.0',
+                      'scipy>=1.4.0',
                       ],
 
     # List additional groups of dependencies here (e.g. development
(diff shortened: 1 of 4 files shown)


schorrm:
I'm sorry, this is getting caught in the crossfire of the rewrite, 🤷‍♂️

TheCleric:
This should be good now @schorrm

setup.py
```diff
@@ -80,8 +80,8 @@
                       'requests>=2.18.1',
                       'lxml>=4.2.1',
                       'pygithub>=1.51',
-                      'altair',
-                      'scipy>=1.4.0'
+                      'altair>=4.1.0',
```
schorrm, line 83: curious where you got the specific version number from

TheCleric: It was what I had installed. We can certainly change this if you'd like. We'd need to identify the earliest working version possibly.

schorrm: Mine works on 4.0.1. @andersonfrailey -- you authored the initial PR for this, do you know what the VN is?

TheCleric: I played around locally with hard coding the version. It seems the earliest one that doesn't crash on the examples is 4.0.0

I can update to that if you'd like.

schorrm: sounds good

PR #128, opened 2020-09-23
Caching!

TheCleric:
This PR adds the ability to cache any function that returns a dataframe object. This cache has been added to all known functions that return a dataframe object (Statcast, Fangraphs, BRef, etc). To add it to a new function:
```python
import pandas as pd

from pybaseball import cache

@cache.dataframe_cache() # This function will be cached when enabled
def new_func() -> pd.DataFrame:
    return pd.DataFrame([])
```

## About this implementation
* Cache is disabled by default
    * Enabling cache:
    ```python
    from pybaseball import cache
    
    cache.enable()
    ```
    * Disabling cache:
    ```python
    from pybaseball import cache
    
    cache.disable()
    ```
    * If caching is ever enabled by default later, caching is purposefully disabled before all unit tests to prevent false results.
* By default it will cache to the `.pybaseball/cached_data` folder in the user's home directory, so cache can be used across projects (directory will be created if not present).
* It supports the following cache storage options: CSV, Parquet
    * Changing the storage mechanism:
    ```python
    from pybaseball import cache
    
    cache.config.cache_type = 'csv'
    ```
    * Default cache type is Parquet.
* Cache directory can also be configured by setting the `PYBASEBALL_CACHE` environment variable to your desired cache directory.
* This will also cache lahman data and moves the directory that `lahman.download_lahman()` places its data to the cache directory.
* This cache is intelligent only at the function parameter level, meaning that calls to the same function with the same params will reuse the same cache value. For simplicity for now, the cache purposefully does not do any subset caching. E.g., a call to `pybaseball.batting_leaders(2000, 2020)`, a follow up call to `pybaseball.batting_leaders(2010, 2015)` will not attempt to reuse the cache, despite likely having the data to do so. This is purposeful for now to get a solution in place that can be improved later.

I consider this PR a chance to discuss the approach and decisions I've made here and to adapt. If large changes are requested though I'd recommend deciding whether what is in this is a good start or to start from scratch. 😄


schorrm:
Do you have PyArrow? My Parquet was much faster.

TheCleric:
> Do you have PyArrow? My Parquet was much faster.

Yes, if you try to run any of the save/load parquet functions without it installed you get:

```
ImportError: Unable to find a usable engine; tried using: 'pyarrow', 'fastparquet'.
A suitable version of pyarrow or fastparquet is required for parquet support.
Trying to import the above resulted in these errors:
 - Missing optional dependency 'pyarrow'. pyarrow is required for parquet support. Use pip or conda to install pyarrow.
 - Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
```

I don't think I had noticed that fastparquet option last time, so I tried that instead.

Somehow that was slower...

12.842 seconds for the run of 10.

However in researching this I noticed that Pandas `to_` and `read_` functions support a variety of compression algorithms and engines depending on the storage type (like CSV vs CSV_GZ). So I've added a lot more options so we can get an idea of where the performance/file size trade offs are.

As well, per our discussion @schorrm, here is a comparison of each method broken down by read speed, write speed, and file size generated. This is for 20 years of batting data (`pybaseball.batting_stats(2000, 2020)`), and only counts the actual read and actual write for 1 call each.o

| Cache Type | Write Speed (seconds) | Read Speed (seconds) | Generated File Size |
| --- | --- | --- | --- |
| CacheType.CSV | 0.819 | 0.138 | 4313.6 KB |
| CacheType.CSV_BZ2 | 0.982 | 0.268 | 998.1 KB |
| CacheType.CSV_GZ | 1.748 | 0.146 | 1409.7 KB |
| CacheType.CSV_XZ | 3.805 | 0.229 | 1095.9 KB |
| CacheType.CSV_ZIP | 1.436 | 0.162 | 1441.5 KB |
| CacheType.EXCEL | 28.763 | 25.116 | 4528.7 KB |
| CacheType.FEATHER | 0.101 | 0.024 | 3009.1 KB |
| CacheType.FEATHER_LZ4 | 0.056 | 0.023 | 3009.1 KB |
| CacheType.FEATHER_UNCOMPRESSED | 0.053 | 0.021 | 8128.7 KB |
| CacheType.FEATHER_ZSTD | 0.057 | 0.023 | 1807.7 KB |
| CacheType.JSON | 0.142 | 0.813 | 11472.6 KB |
| CacheType.JSON_BZ2 | 1.000 | 0.947 | 2201.4 KB |
| CacheType.JSON_GZ | 3.113 | 0.524 | 3323.7 KB |
| CacheType.JSON_XZ | 6.793 | 0.685 | 1542.1 KB |
| CacheType.JSON_ZIP | 0.798 | 0.524 | 3344.3 KB |
| CacheType.PARQUET | 0.146 | 0.153 | 1604.5 KB |
| CacheType.PARQUET_BROTLI | 0.125 | 0.067 | 1604.5 KB |
| CacheType.PARQUET_GZ | 0.179 | 0.070 | 1194.6 KB |
| CacheType.PARQUET_SNAPPY | 0.135 | 0.065 | 1333.0 KB |
| CacheType.PARQUET_FAST | 2.044 | 2.264 | 5724.2 KB |
| CacheType.PARQUET_FAST_BROTLI | 0.270 | 0.115 | 5724.2 KB |
| CacheType.PARQUET_FAST_GZ | 0.490 | 0.150 | 1359.6 KB |
| CacheType.PARQUET_FAST_SNAPPY | 0.303 | 0.113 | 2070.5 KB |
| CacheType.PICKLE | 0.056 | 0.017 | 8018.7 KB |
| CacheType.PICKLE_BZIP | 0.054 | 0.021 | 8018.7 KB |
| CacheType.PICKLE_GZ | 2.852 | 0.035 | 1265.5 KB |
| CacheType.PICKLE_XZ | 2.804 | 0.078 | 875.3 KB |
| CacheType.PICKLE_ZIP | 0.301 | 0.028 | 1318.1 KB |

TheCleric:
For fun, I also did a little but of "ranking" this data where I put it in a dataframe, and gave each method 0-100 points for read speed, write speed, and file size. I then averaged the read/write speed points with the file size points, and here's how it looks. Just depends how you would want to weight disk IO vs file size:

```python
import pandas as pd
raw = [
    ['CSV', 0.819, 0.138, 4313.6],
    ['CSV_BZ2', 0.982, 0.268, 998.1],
    ['CSV_GZ', 1.748, 0.146, 1409.7],
    ['CSV_XZ', 3.805, 0.229, 1095.9],
    ['CSV_ZIP', 1.436, 0.162, 1441.5],
    ['EXCEL', 28.763, 25.116, 4528.7],
    ['FEATHER', 0.101, 0.024, 3009.1],
    ['FEATHER_LZ4', 0.056, 0.023, 3009.1],
    ['FEATHER_UNCOMPRESSED', 0.053, 0.021, 8128.7],
    ['FEATHER_ZSTD', 0.057, 0.023, 1807.7],
    ['JSON', 0.142, 0.813, 11472.6],
    ['JSON_BZ2', 1.000, 0.947, 2201.4],
    ['JSON_GZ', 3.113, 0.524, 3323.7],
    ['JSON_XZ', 6.793, 0.685, 1542.1],
    ['JSON_ZIP', 0.798, 0.524, 3344.3],
    ['PARQUET', 0.146, 0.153, 1604.5],
    ['PARQUET_BROTLI', 0.125, 0.067, 1604.5],
    ['PARQUET_GZ', 0.179, 0.070, 1194.6],
    ['PARQUET_SNAPPY', 0.135, 0.065, 1333.0],
    ['PARQUET_FAST', 2.044, 2.264, 5724.2],
    ['PARQUET_FAST_BROTLI', 0.270, 0.115, 5724.2],
]

df = pd.DataFrame(data=raw, columns=['CacheType', 'WriteSpeed', 'ReadSpeed', 'FileSize'])

ticks = 100

write_speed_floor = df['WriteSpeed'].min()
write_speed_ceiling = df['WriteSpeed'].max()
write_speed_increment = (write_speed_ceiling - write_speed_floor) / ticks

df['WriteSpeedRating'] = ticks - (df['WriteSpeed'] - write_speed_floor) / write_speed_increment

read_speed_floor = df['ReadSpeed'].min()
read_speed_ceiling = df['ReadSpeed'].max()
read_speed_increment = (read_speed_ceiling - read_speed_floor) / ticks

df['ReadSpeedRating'] = ticks - (df['ReadSpeed'] - read_speed_floor) / read_speed_increment

file_size_speed_floor = df['FileSize'].min()
file_size_speed_ceiling = df['FileSize'].max()
file_size_speed_increment = (file_size_speed_ceiling - file_size_speed_floor) / ticks

df['FileSizeRating'] = ticks - (df['FileSize'] - file_size_speed_floor) / file_size_speed_increment

df['OverallRating'] = ((df['WriteSpeedRating'] + df['ReadSpeedRating']) / 2 + df['FileSizeRating']) / 2

df.sort_values('OverallRating', ascending=False)
```


CacheType | WriteSpeed | ReadSpeed | FileSize | WriteSpeedRating | ReadSpeedRating | FileSizeRating | OverallRating
-- | -- | -- | -- | -- | -- | -- | --
CSV_BZ2 | 0.982 | 0.268 | 998.1 | 96.764194 | 99.015740 | 100.000000 | 98.944983
PARQUET_GZ | 0.179 | 0.070 | 1194.6 | 99.561129 | 99.804742 | 98.124015 | 98.903475
PARQUET_SNAPPY | 0.135 | 0.065 | 1333.0 | 99.714385 | 99.824666 | 96.802711 | 98.286119
PARQUET_BROTLI | 0.125 | 0.067 | 1604.5 | 99.749216 | 99.816697 | 94.210702 | 96.996829
PARQUET | 0.146 | 0.153 | 1604.5 | 99.676071 | 99.473999 | 94.210702 | 96.892869
CSV_ZIP | 1.436 | 0.162 | 1441.5 | 95.182863 | 99.438135 | 95.766862 | 96.538681
CSV_GZ | 1.748 | 0.146 | 1409.7 | 94.096134 | 99.501893 | 96.070457 | 96.434735
FEATHER_ZSTD | 0.057 | 0.023 | 1807.7 | 99.986068 | 99.992030 | 92.270753 | 96.129901
CSV_XZ | 3.805 | 0.229 | 1095.9 | 86.931383 | 99.171150 | 99.066304 | 96.058785
JSON_BZ2 | 1.000 | 0.947 | 2201.4 | 96.701498 | 96.310022 | 88.512101 | 92.508930
JSON_XZ | 6.793 | 0.685 | 1542.1 | 76.523859 | 97.354055 | 94.806435 | 90.872696
FEATHER_LZ4 | 0.056 | 0.023 | 3009.1 | 99.989551 | 99.992030 | 80.800993 | 90.395892
FEATHER | 0.101 | 0.024 | 3009.1 | 99.832811 | 99.988045 | 80.800993 | 90.355711
JSON_ZIP | 0.798 | 0.524 | 3344.3 | 97.405085 | 97.995617 | 77.600840 | 87.650596
JSON_GZ | 3.113 | 0.524 | 3323.7 | 89.341693 | 97.995617 | 77.797508 | 85.733081
CSV | 0.819 | 0.138 | 4313.6 | 97.331940 | 99.533772 | 68.346938 | 83.389897
PARQUET_FAST_BROTLI | 0.270 | 0.115 | 5724.2 | 99.244166 | 99.625423 | 54.879947 | 77.157371
PARQUET_FAST | 2.044 | 2.264 | 5724.2 | 93.065134 | 91.061965 | 54.879947 | 73.471748
FEATHER_UNCOMPRESSED | 0.053 | 0.021 | 8128.7 | 100.000000 | 100.000000 | 31.924197 | 65.962098
JSON | 0.142 | 0.813 | 11472.6 | 99.690003 | 96.843993 | 0.000000 | 49.133499
EXCEL | 28.763 | 25.116 | 4528.7 | 0.000000 | 0.000000 | 66.293379 | 33.146690

pybaseball/datahelpers/caching.py
```diff
@@ -0,0 +1,313 @@
+import copy
```
schorrm, line 1: This file is too much.

pybaseball/lahman.py
```diff
@@ -1,9 +1,12 @@
...
-import pandas as pd
 from io import BytesIO
+from os import path
 from typing import Optional
+from zipfile import ZipFile
+
+import pandas as pd
+import requests
+
+from .datahelpers import caching
```
schorrm, line 9: I think Lahman should be treated separately, as a properly stable database.

schorrm:
Overall notes: this is too complicated. Much too complicated.

TheCleric:
> Overall notes: this is too complicated. Much too complicated.

What about it do you find complicated? I agree that I've added WAY too many options for caching mechanisms, but that was simply so we can identify and evaluate which seem beneficial. Those could be trimmed down. Other than that I think it strikes a good balance of flexibility and complexity.

pybaseball/datahelpers/caching.py
TheCleric: In what way?

pybaseball/lahman.py
TheCleric: I removed the one cache wrapper in here, but think keeping the use of the caching directory here makes sense.

schorrm:
Henry Ford, black, we're gonna pick our black.

pybaseball/datahelpers/caching.py
schorrm: 100 LOC target

TheCleric:
> Henry Ford, black, we're gonna pick our black.

TheCleric:
Whoops, didn't mean to close!

TheCleric:
> Henry Ford, black, we're gonna pick our black.

I agree we can simplify a bit in terms of catching methos, but I also hate when a library makes assumptions on my behalf with no way to disagree with them.

pybaseball/datahelpers/caching.py
TheCleric: That seems like a very aggressive target for a subsystem like this one. I'll see what I can do about splitting it up, but as a hard limit that may be difficult.

schorrm: I'll try and take a knife to it

pybaseball/lahman.py
schorrm: I think we should use the `download_lahman`, but point to `.pybaseball/` so it ends up like `.pybaseball/baseball-databank` or whatever it is

schorrm:
Fair, but the current system is definitely wrong.
My current feelings: allow three options.
1. Parquet (default). I still think based on my testing that it has the best overall combo -- it's the best on space and it's also competitive on time.
2. CSV. Poor format on all metrics, but readable. Available.
3. Pickle. I don't know who would need to load the cache in 20ms instead of 50ms. If you are that person, pickle is the option for you.

README.md
```diff
@@ -215,6 +215,59 @@ This function returns a list of dataframes. Each dataframe is the standings for
...
+caching option like so:
+
+```python
+from pybaseball.datahelpers import caching
+
+caching.cache_config.enable()
+```
+
+This will store a copy of the data returned for each DataFrame returning function call.
+This data will be stored in `~/.pybaseball/cached_data` (or `%USERPROFILE%\.pybaseball\cached_data` in Windows) by
```
schorrm, line 234: Just `.pybaseball/cache` will do I think

pybaseball/datahelpers/caching/cache_type.py
```diff
@@ -0,0 +1,27 @@
...
+    JSON_ZIP = 'json.zip'
+    PARQUET = 'parquet'
+    PARQUET_GZ = 'gz.parquet'
+    PARQUET_FAST = 'fast.parquet'
+    PARQUET_FAST_GZ = 'fast.gz.parquet'
+    PICKLE = 'pickle'
+    PICKLE_BZIP = 'pickle.bzip'
+    PICKLE_GZ = 'pickle.gz'
+    PICKLE_XZ = 'pickle.xz'
+    PICKLE_ZIP = 'pickle.zip'
```
schorrm, line 27: Unnecessary. Maybe good for evaluation, but as I said, let's take our limited set and roll with it.

pybaseball/lahman.py
```diff
@@ -26,15 +29,20 @@ def download_lahman():
...
         z = get_lahman_zip()
         # this way we'll now start using the extracted zip directory
         # instead of the session ZipFile object
 
 def _get_file(tablename: str, quotechar: str = "'") -> pd.DataFrame:
     z = get_lahman_zip()
     f = f'{base_string}/{tablename}'
-    data = pd.read_csv(f if z is None else z.open(f), header=0, sep=',', quotechar=quotechar)
+    data = pd.read_csv(
+        f"{path.join(caching.cache_config.cache_directory, f)}" if z is None else z.open(f),
```
schorrm, line 41: Do this on `base_string` instead I think

pybaseball/league_pitching_stats.py
```diff
@@ -1,9 +1,13 @@
...
 import io
+
+import numpy as np
+import pandas as pd
+import requests
 from bs4 import BeautifulSoup
-import datetime
+
+from .datahelpers import caching
+
```
schorrm, line 10: Do you have to reorder these all the time? Adds to the diff and not a really good reason for it

pybaseball/playerid_lookup.py
```diff
@@ -1,19 +1,26 @@
...
+
+@caching.dataframe_cache()
 def chadwick_register(save: bool = False) -> pd.DataFrame:
     ''' Get the Chadwick register Database '''
 
-    if os.path.exists(register_file):
-        table = pd.read_csv(register_file)
+    if os.path.exists(get_register_file()):
+        table = pd.read_csv(get_register_file())
         return table
```
schorrm, line 24: I think this makes sense separately. Perhaps, even pull from the GitHub?

pybaseball/playerid_lookup.py
```diff
@@ -1,19 +1,26 @@
...
+from .datahelpers import caching
+
 # dropped key_uuid. looks like a has we wouldn't need for anything.
 # TODO: allow for typos. String similarity?
 
 url = "https://raw.githubusercontent.com/chadwickbureau/register/master/data/people.csv"
-register_file = 'chadwick-register.csv'
 
+def get_register_file():
+    return os.path.join(caching.cache_config.cache_directory, 'chadwick-register.csv')
```
schorrm, line 15: Why does this need to be a function?

setup.py
```diff
@@ -79,6 +79,8 @@
                       'beautifulsoup4>=4.4.0',
                       'requests>=2.18.1',
                       'lxml>=4.2.1',
+                      'fastparquet>=0.4.1',
```
schorrm, line 82: remove fastparquet. Pyarrow alone should be sufficient (as the faster of the two)

tests/integration/pybaseball/datahelpers/test_caching.py
```diff
@@ -0,0 +1,30 @@
...
+                                                          cache_directory=path.join(pybaseball.datahelpers.caching.cache_config.cache_directory, '.pytest'))
+    ):
+        # Delete any existing data just in case
+        pybaseball.datahelpers.caching.bust_cache()
+
+        # Uncached read, cached write
+        result, _, write_time, _ = pybaseball.batting_stats(2019) # type: ignore
+
+        # Cached read, no write
+        result2, read_time2, _, file_size2 = pybaseball.batting_stats(2019) # type: ignore
```
schorrm, line 23: Where did the other parameters come from?

tests/integration/pybaseball/datahelpers/test_caching.py
```diff
@@ -0,0 +1,30 @@
+from unittest.mock import patch
```
schorrm, line 1: This entire file should be gone in the final version I think

tests/pybaseball/datahelpers/caching/test_caching.py
```diff
@@ -0,0 +1,344 @@
...
+def _mock_data_1() -> pd.DataFrame:
+    return pd.DataFrame([1, 2], columns=['a'])
+
+
+@pytest.fixture(name="mock_data_2")
+def _mock_data_2() -> pd.DataFrame:
+    return pd.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
+
+
+class TestCacheWrapper:
```
schorrm, line 32: Can this go on a diet? This is really long and I don't know what it does (which is a recurring problem across our unit testing honestly)

.github/workflows/python-publish.yml
```diff
@@ -9,23 +9,22 @@ on:
```
schorrm, line 9: This file does not belong here at all

README.md
```diff
@@ -215,6 +215,59 @@ This function returns a list of dataframes. Each dataframe is the standings for
...
+By default caching is disabled so as to respect a user's potential desire to not have their hard drive space used
+without their permission. However, enabling caching is simple.
+
+Caching can be turned on by including the caching module in the pybaseball.datahelpers submodule, and enabling the
+caching option like so:
+
+```python
+from pybaseball.datahelpers import caching
+
+caching.cache_config.enable()
```
schorrm, line 230: I would prefer
```python3
import pybaseball
pybaseball.cache.enable()
```
and correspondingly
```python3
pybaseball.cache.disable()
```

docs/caching.md
```diff
@@ -0,0 +1,57 @@
...
+    * Cache speed tests. Each represents an initial load of 5 different functions, and then 999 cached loads:
+        * CSV: 118.940 seconds
+        * CSV_GZ: 127.987 seconds
+        * EXCEL: Not tested based off the 10 iteration result
+        * FEATHER: 42.061 seconds
+        * JSON: 432.338 seconds
+        * PARQUET: 95.881 seconds
+        * PICKLE: 22.616 seconds
+    * Default cache type is CSV.
+* Expirations can be set on a cache, where cache files past a certain timedelata will be ignored/discarded:
```
schorrm, line 42: These need to go

pybaseball/datahelpers/caching/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+                return result, read_time, write_time, os.path.getsize(filename)
+            else:
+                return result
+
+        if self.reset_cache:
+            self.bust_cache(func)
+
+        return _cached
+
+    def bust_cache(self, func: Callable) -> None:
```
schorrm, line 98: Rename to `flush_cache` I think

schorrm: Or, even:
`pybaseball.cache.flush()`

pybaseball/datahelpers/caching/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+        cache_files = [
+            x for x in os.walk(self.cache_config.cache_directory).__next__()[2] if x.startswith(f"{func.__name__}(")
```
schorrm, line 100: ????

pybaseball/datahelpers/caching/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def save(self, data: pd.DataFrame, filename: str) -> None:
+        dataframe_utils.save(data, filename, self.cache_config.cache_type)
+
+def bust_cache() -> None:
+    try:
+        _rmtree(cache_config.cache_directory)
+    except:
+        pass
+    _mkdir(cache_config.cache_directory)
```
schorrm, line 117: why not just
```python3
for f in os.listdir(...):
    os.remove(f)
```

pybaseball/amateur_draft.py
```diff
@@ -1,12 +1,16 @@
...
+
+
+def get_draft_results(year, round):
     url = f"https://www.baseball-reference.com/draft/?year_ID={year}&draft_round={round}&draft_type=junreg&query_type=year_round&"
-    res = requests.get(url, timeout=None).content 
+    res = requests.get(url, timeout=None).content
     draft_results = pd.read_html(res)
     return draft_results
 
+@caching.dataframe_cache()
```
schorrm, line 13: Can we have a shorter decorator if it's the only one?
Just `@caching.cache()`

pybaseball/datahelpers/caching/dataframe_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    elif cache_type in [CacheType.PICKLE, CacheType.PICKLE_BZIP, CacheType.PICKLE_GZ, CacheType.PICKLE_XZ, CacheType.PICKLE_ZIP]:
+        data = pd.read_pickle(filename)
+    else:
+        raise ValueError(f"cache_type of {cache_type} is unsupported.")
+    data.drop(
+        data.columns[data.columns.str.contains('unnamed', case=False)],
+        axis=1,
+        inplace=True
+    )
+    return data
```
schorrm, line 28: Snoop it off the extension. I think:
```python3
ext_to_func = {'csv' : pd.read_csv, 'pickle': pd.read_pickle, 'parquet': pd.read_parquet}
def load(filename):
  ext = filename.split('.')[-1]
  return ext_to_func[ext](filename)
```

pybaseball/datahelpers/caching/func_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+    args_key = ', '.join([args_hash, kwargs_hash]).strip(', ')
+
+    # If the args_key is very long, just use a sha hash
+
+    if len(args_key) > MAX_ARGS_KEY_LENGTH:
+        # Let's hash this to make the filename shorter
+        args_key = hashlib.sha256(args_key.encode('utf-8')).hexdigest()
+
+    return f"{f_hash}({args_key})"
```
schorrm, line 71: why are we rolling our own hash?

tests/pybaseball/datahelpers/caching/test_func_utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_get_func_hash_long_path_windows(self) -> None:
+        def test_func(*args: Any, **kwargs: Any) -> None:
+            pass
+
+        expected_hash = "test_func(93ece9cbe326c434412e28b87d466bd08bc0f2e7db0770cff852e7b9f46240d8)"
+
+        long_kwargs = {f"val{str(x)}": x for x in range(caching.MAX_ARGS_KEY_LENGTH)}
+
+        assert caching.get_func_hash(test_func, ('a', 1), long_kwargs) == expected_hash
```
schorrm, line 122: Why did we roll our own hash function to begin with, and, given that, why do we have to have 122 lines of unit testing for it?

schorrm (changes requested):

pybaseball/datahelpers/caching/cache_type.py
TheCleric: > Unnecessary. Maybe good for evaluation, but as I said, let's take our limited set and roll with it.

Yeah I agree. To be clear this was just to evaluate our options. I didn't intend to merge with all these in place. 😀

pybaseball/lahman.py
TheCleric: Due to the fact that the base string is the relative path in the zip file AND on the local file system, I don't believe you can.

pybaseball/league_pitching_stats.py
TheCleric: I reorder them in the standard python ordering. From PEP8 (https://www.python.org/dev/peps/pep-0008/#imports)

Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.

pybaseball/playerid_lookup.py
TheCleric: Possibly as a future improvement. For now adding a decent cache with two lines of code seems like a win.

pybaseball/playerid_lookup.py
TheCleric: Because the cache_directory is user modifiable. If the user changes it after this file is imported, this would have the wrong value.

setup.py
TheCleric: Agreed.

tests/integration/pybaseball/datahelpers/test_caching.py
TheCleric: If profiling is turned on, the cache wrapper will return these.

tests/integration/pybaseball/datahelpers/test_caching.py
TheCleric: Vehemently and completely disagree. This ensures that the cache does what it is intended to do, and that any future modifications can be done safely and without regressions. The code code coverage on the cache has 100% coverage, so every line is under test. Why would we want to NOT test code?

tests/pybaseball/datahelpers/caching/test_caching.py
TheCleric: Its really long hecause it tests the entire cache thoroughly. So no it cannot go on a diet unless we want to have untested code scenarios, which I can't even imagine a compelling reason for.

pybaseball/datahelpers/caching/__init__.py
TheCleric: Because os.remove doe not delete subdirectories (such as lahman, or any future needs).

pybaseball/amateur_draft.py
TheCleric: I think being explicit here is good, especially if we add any other cache types later.

.github/workflows/python-publish.yml
TheCleric: It did at the time because I had to modify it to get the snappy compression to work. I removed that as it was problematic, so this is just the formatting that was done. I'll reset this one.

pybaseball/amateur_draft.py
schorrm: If we add later, than rename it I think.

schorrm: even then, just to `df_cache`

pybaseball/datahelpers/caching/__init__.py
schorrm: Lahman should be kept separately, not in the `cache/` subdir

docs/caching.md
TheCleric: The speed tests?

pybaseball/datahelpers/caching/__init__.py
TheCleric: @schorrm this comment is not helpful.

pybaseball/amateur_draft.py
TheCleric: Why? To save 7 characters?

pybaseball/datahelpers/caching/func_utils.py
TheCleric: Because it generates something human readable if possible (such as `func_name(2019, keyword_param_name=True)`) so that the cache files can be viewed easily on the drive.

tests/pybaseball/datahelpers/caching/test_func_utils.py
TheCleric: See above.

pybaseball/datahelpers/caching/__init__.py
TheCleric: So we force the user to take that, with no option to change?

pybaseball/datahelpers/caching/dataframe_utils.py
TheCleric: That could be possible once we decide on the final caching options are. I'd like to wait for other input to make sure we make the right decision there.

pybaseball/datahelpers/caching/func_utils.py
schorrm: Can we reduce this to along the lines
```python3
def get_func_hash(func: Callable, args: Tuple, kwargs: Dict) -> str:
    fname = get_func_name(func)
    args_s = str(args)
    kwargs_s = str(kwargs)
    args_key = f'{{args_s}--{kwargs_s}'
    # If the args_key is very long, just use a sha hash

    if len(args_key) > MAX_ARGS_KEY_LENGTH:
        # Let's hash this to make the filename shorter
        args_key = hashlib.sha1(args_key.encode('utf-8')).hexdigest()

    return f"{f_name}({args_key})"
```

docs/caching.md
schorrm: Useful for choosing on the GitHub, not suitable for user-facing documentation I think

pybaseball/amateur_draft.py
schorrm: cleaner. I prefer `@caching.cache()`

README.md
TheCleric: Done

pybaseball/datahelpers/caching/__init__.py
schorrm: Walk next 2? It's unclear what this does / how?

pybaseball/datahelpers/caching/__init__.py
schorrm: Lahman is stable. It's about a couple updates over the winter. It's not volatile the same way.

pybaseball/league_pitching_stats.py
schorrm: Good call, if you're already adding something there.

pybaseball/playerid_lookup.py
schorrm: Makes sense.

tests/integration/pybaseball/datahelpers/test_caching.py
schorrm: Minimize the CacheType thingy?

pybaseball/datahelpers/caching.py
schorrm: I don't want to split it up, I want to put it on a diet.

pybaseball/datahelpers/caching/__init__.py
schorrm: And yeah, for this version, we force it (we already force the name in `download_lahman`), and we can throw in something else later. For simplicity's sake.

README.md
TheCleric: Done. As much as I like having things categorized (e.g., datahelpers.caching), this does look cleaner.

docs/caching.md
TheCleric: Completely fair point. Just want to make sure I understood what was meant here by "these need to go".

pybaseball/datahelpers/caching/__init__.py
TheCleric: Okay, I'll break this down a bit to be clearer about what it is doing.

TheCleric: Actually since this should be flat I'll switch this to `os.listdir`

TheCleric: Done

pybaseball/datahelpers/caching/func_utils.py
TheCleric: No for a few reasons:

1. If any of the args or kwargs has a character that does not belong in a filename (such as `/`)
2. `str(value)` is not guaranteed to be unique across objects. So a function call may inadvertently return a cached result despite never being called with the params it received.

Hence why I wrote the `get_value_hash` function to ensure safe parsing.

tests/integration/pybaseball/datahelpers/test_caching.py
TheCleric: Yeah that's definitely going to happen, just waiting for everyone to get their feedback in so that we can make sure we have the correct, minimal cache options.

pybaseball/datahelpers/caching/__init__.py
TheCleric: > And yeah, for this version, we force it..and we can throw in something else later.

Do we have a compelling reason to do that? The code is written, is beneficial to the user, and works. Why throw the code away and then have to do it again later?

pybaseball/datahelpers/caching/__init__.py
TheCleric: Done.

pybaseball/amateur_draft.py
TheCleric: I prefer being explicit over implicit here.

pybaseball/datahelpers/caching/func_utils.py
schorrm: Can you elaborate on the edge case in (2)? Because I would really like to try and figure out how we can avoid this system

TheCleric: Sure. 2 is actually two separate issues. I'll give this code example to illustrate both:

```python
class A:
    def __init__(self, a, b):
        self.a = a
        self.b = b

x = A(1, 2)
print(str(x))

xy= A(1, 2)
print(str(y))
```

On my machine I get:
```
<__main__.A object at 0x000001C0435BEAC0>
<__main__.A object at 0x000001C0435BE340>
```

However, rerunning it, I get two entirely different results. So here's why that is important to us:

1. If another object generates that same value at some point (highly unlikely in this example, but not impossible, also is HIGHLY dependent on the implementation of `__str__` or `__repr__` for the value being passed to `str`).
2. More importantly, in this example we would want both `x` and `y` to be equivalent to our cache, but using `str` would mean we'd skip the cache as we'd see these as different values.

Now let's compare to what we generate (after I fixed a bug, and added a test to ensure this is true):
```
A({a=1, b=2})
A({a=1, b=2})
```

So this would generate a cache hit for each of those objects.

pybaseball/datahelpers/caching/cache_type.py
TheCleric: This has now been cut back to CSV, PARQUET, and PICKLE with PARQUET as the default

pybaseball/datahelpers/caching/func_utils.py
schorrm: if you noticed, I left in
```python3
    fname = get_func_name(func)
```
Since that object issue is important for functions.
However: do we use pure object code there? Do we even have anything along the lines of A() going into the hash?
As I see it, there are three components:
1. The function. `get_func_name` solves that string.
2. The args. Should be a list? Should be stringable in turn then, no?
3. The kwargs. Should be a dict, should be stringable too, no?

pybaseball/cache/cache_config.py
```diff
@@ -0,0 +1,25 @@
+import pathlib
+from datetime import timedelta
+from os import path
+
+from .cache_type import CacheType
+from .file_utils import _mkdir
+
+class CacheConfig:
```
schorrm, line 8: Dumb question: where does this get stored?

README.md
```diff
@@ -215,6 +215,55 @@ This function returns a list of dataframes. Each dataframe is the standings for
...
+```
+
+Each CacheConfig option is as follows:
+- `enabled` - bool - Whether cache should be enabled
+- `cache_directory` - str - The location to store the cached data. If it does not exist, it will be created.
+- `expiration` - datetime.timedelta - The timedelta after the cache file is created to expire it. Default = 24 hours.
+- `cache_type` - pybaseball.cache.CacheType - The method to use in storing the cache. Options:
+  - `CacheType.CSV` - Cache is stored in pandas compatible CSV format files
+  - `CacheType.PARQUET` - Cache is stored in Apache Parquet format files: https://parquet.apache.org/
+  - `CacheType.PICKLE` - Cache is stored as python pickle files
```
schorrm, line 264: We really don't need an enum for 3 values

schorrm: or 2

pybaseball/cache/cache_config.py
TheCleric: It's in memory only:

See line 17 of https://github.com/jldbc/pybaseball/pull/128/files#diff-5566a5251165c3b840f204a66a4506a3

README.md
TheCleric: Makes sense. I'll tweak.

TheCleric: And now down to CSV and PARQUET, and those are just strings (validated with a Literal type by mypy).

TheCleric:
FYI tests are failing here due to a bad test in master fixed in #136

pybaseball/datahelpers/caching/func_utils.py
TheCleric: @schorrm this is a fair point. Looking through the existing functions we'll be caching, all of them should be stringable (no classes).

At this point it's a matter of do we want the simple solution, or the future proof solution (in case we do want to cache a function that would take an object)? I could see an argument for either path.

I will say we would have to do something a little more complex than you suggested as we'll have to make sure there are no reserved filename characters in the generated file path.

TheCleric:
@schorrm this now has your changes integrated in. I made the following tweaks:

- Fix the tests
- Fix some of the type annotations
- Changed some namespacing around to avoid namespace collisions
- Convert CacheConfig to a singleton type object so that multiple modules can share the same config
- Fix a bug I found (from my original code) with the CSV loader.

setup.py
```diff
@@ -79,9 +79,11 @@
                       'beautifulsoup4>=4.4.0',
                       'requests>=2.18.1',
                       'lxml>=4.2.1',
+                      'pyarrow>=1.0.1',
                       'pygithub>=1.51',
                       'altair>=4.0.0',
                       'scipy>=1.4.0',
+                      'openpyxl>=3.0.5',
```
schorrm, line 86: Why?

pybaseball/league_batting_stats.py
```diff
@@ -42,7 +45,7 @@ def get_soup(start_dt, end_dt):
     # if((start_dt is None) or (end_dt is None)):
     #    print('Error: a date range needs to be specified')
     #    return None
-    url = "http://www.baseball-reference.com/leagues/daily.cgi?user_team=&bust_cache=&type=b&lastndays=7&dates=fromandto&fromandto={}.{}&level=mlb&franch=&stat=&stat_value=0".format(start_dt, end_dt)
+    url = "http://www.baseball-reference.com/leagues/daily.cgi?user_team=&flush_cache=&type=b&lastndays=7&dates=fromandto&fromandto={}.{}&level=mlb&franch=&stat=&stat_value=0".format(start_dt, end_dt)
```
schorrm, line 48: whoops

schorrm: occurs for other BRef URLs too

pybaseball/cache/cache_config.py
```diff
@@ -0,0 +1,74 @@
...
+    def __new__(cls, enabled: bool = False, cache_directory: str = None, default_expiration: int = None,
+                cache_type: Optional[str] = None) -> 'CacheConfig':
+        if not CacheConfig.__INSTANCE__:
+            CacheConfig.__INSTANCE__ = super(CacheConfig, cls).__new__(cls)
+
+        CacheConfig.__INSTANCE__._set(enabled, cache_directory, default_expiration, cache_type)
+        return CacheConfig.__INSTANCE__  # type: ignore
+
+    def _set(self, enabled: bool = False, cache_directory: str = None, default_expiration: int = None,
+             cache_type: Optional[str] = None) -> None:
```
schorrm, line 35: I think we should remove `cache_directory` from the API. This is on me really.

setup.py
TheCleric: Snuck in combining the branches. Will fix.

pybaseball/league_batting_stats.py
TheCleric: Ugh. Will fix.

TheCleric: Thanks for catching this.

pybaseball/cache/cache_config.py
TheCleric: I have a fallback idea. I'll add it and see what you think.

TheCleric: Okay I removed cache_directory as a value in the new/set for this object. If the user wants to override I've added an environment variable.

tests/pybaseball/cache/test_cache_config.py
```diff
@@ -33,9 +33,9 @@ def test_cache_directory_default(mkdir: MagicMock) -> None:
 
 def test_cache_directory_set(mkdir: MagicMock) -> None:
     my_dir: str = '~/my_dir'
-    config = cache.CacheConfig(cache_directory=my_dir)
+    cache.config.cache_directory = my_dir
 
-    assert config.cache_directory == my_dir
+    assert cache.config.cache_directory == my_dir
     assert mkdir.called_once_with(my_dir)
```
schorrm: Is this valid usage at all now?

schorrm:
I should have suggested this earlier, mea culpa:
What do you think of dropping all the actual usage of the decorator? It'll minimize diff / merge conflicts, and anyway we're going to want to have the freshness in there, I think?
Also, should we allow some sort of option to 'next season', e.g. keep until the next March 1st?

schorrm:
LGTM. I'd like to strip the dev artifacts now and then commit.

tests/pybaseball/cache/test_cache_config.py
TheCleric: It's technically allowed (as it's being set within the object). Whether it should be or not is certainly up for debate. I'd probably argue no.

TheCleric:
> I should have suggested this earlier, mea culpa:
> What do you think of dropping all the actual usage of the decorator? It'll minimize diff / merge conflicts, and anyway we're going to want to have the freshness in there, I think?

We certainly could do that, but no matter when we add it, we have the potential for merge conflicts. I'd say bite the bullet now and get it over with.

schorrm:
I hear, although we haven't rolled it out intelligently (differential freshness etc)

schorrm:
I'm half tempted to put in a `set_expiry` param in the **call** kwargs, which would be intercepted by the cache caller

TheCleric:
> I'm half tempted to put in a `set_expiry` param in the **call** kwargs, which would be intercepted by the cache caller

I think that would likely lead to a lot of user confusion around why we're passing in a kwarg into a function that doesn't use it (and would certainly cause MyPy validation errors).

TheCleric:
> I hear, although we haven't rolled it out intelligently (differential freshness etc)

Don't let the perfect be the enemy of the good. I'd say this is an improvement over what was there. It itself can be improved later.

schorrm:
Fair enough. I just want to add df_cache as a valid usage

schorrm:
This runs into some merge issues because of the hotfixes. Can we resolve and swap the dataframe_cache to df_cache? After that, literally everything is covered I think

TheCleric:
> This runs into some merge issues because of the hotfixes. Can we resolve and swap the dataframe_cache to df_cache? After that, literally everything is covered I think

Done

schorrm:
Great. Good riddance, this thing.

PR #130, opened 2020-09-24
Update FanGraphs docs, and improve typing in FG code

TheCleric:


docs/fangraphs.md
```diff
@@ -0,0 +1,61 @@
...
+fangraphs.fg_team_fielding_data(2019)
+```
+
+## Team Pitching Stats
+
+```python
+from pybaseball.datasources import fangraphs
+
+fangraphs.fg_team_pitching_data(2019)
+```
```
schorrm, line 61: Combine to one code block

schorrm: And can we use the top level usage here as the example?

docs/fangraphs.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Team Batting Stats
+fangraphs.fg_team_batting_data(2019)
+
+# Team Fielding Stats
+fangraphs.fg_team_fielding_data(2019)
+
+# Team Pitching Stats
+fangraphs.fg_team_pitching_data(2019)
+```
```
schorrm, line 42: I prefer `pybaseball.fg_batting_stats` as more continuous with the established API

TheCleric: Do you want me to add extra scope to this PR to fix that then? Or add a separate PR for that.

schorrm: I thought this was the case prior honestly... I guess you could just throw it in here

schorrm: Hmm... this is already the case actually, checking out the `__init__.py`. I'm gonna modify this so I can merge now

schorrm:
Went thru the CLI due to the conflicts.

PR #132, opened 2020-09-25
Fix issues with standings bref api; add integration tests

TheCleric:
This PR accomplishes two things:

1. Adds integration tests to the standings function, which was our code with the lowest coverage (only 11% under test).
2. While adding those integration tests, I discovered that due to an issue with BRef, all pre-1969 standings were erroring out.  As well I discovered that for 1981, we weren't accurately pulling the standings due to a quirk in BRef. So this fixes those issues as well.

So hooray for tests! They help us to catch bugs before our users do.

### DIFF
--- a/pybaseball/standings.py
+++ b/pybaseball/standings.py
@@ -1,5 +1,4 @@
 import datetime
-import re
 
 import pandas as pd
 import requests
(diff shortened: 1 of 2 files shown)


schorrm:
LGTM

PR #135, opened 2020-09-27
Fix the Fangraphs docs accuracy

TheCleric:


TheCleric:
Broken tests on this PR came from a bad test in master. Will be fixed via #136

schorrm:
Can you rename this something more accurate?

TheCleric:
> Can you rename this something more accurate?

It SHOULD be accurate. I'm still getting changes sneaking in when I rebase, which is REALLY strange, as I know master has the extraneous marcels changes that are showing up here. Let me try to fix...

TheCleric:
MUCH better!

schorrm:
I meant rename the PR

TheCleric:
> I meant rename the PR

I know. The PR is only about fixing the accuracy of the fangraphs.md file.

schorrm:
Ah, ok, because the diff looked like Marcels mainly last time I looked.

PR #137, opened 2020-09-27
Add split stats support from bbref

mwisnie5:
This PR adds a function that returns a dataframe of either batting or pitching splits given a player id. Additionally player bio info such as position, batting and throwing handedness, height, weight, and team can be returned as well. 

Note that this replaces an older PR that required multiple calls to bbref for each split category. This new implementation pulls all the data with a single request.


TheCleric:
The tests on this are failing, but the test that is currently failing will be fixed in #136

docs/split_stats.md
```diff
@@ -0,0 +1,34 @@
...
+`get_splits(playerid, year=None, player_info=False, pitching_splits=False)`
+
+Look up a player's split stats from baseball-reference. Either batting or pitching splits can be returned and splits can be provided for any season or for career. 
+Split stats are returned as a multi-index dataframe by split category and split. 
+Additionaly, player info can also be returned as a dictionary since the information is available without requiring a separate request to the baseball-reference page. 
+
+## Arguments
+`playerid:` String. The player's bbref playerid. Example: Mike Trout is 'troutmi01'
+
+`year:` String. Optional. The year to get split stats for. Leaving this out will provide career split stats.
```
TheCleric, line 12: This should probably be an int, not a string.

pybaseball/split_stats.py
```diff
@@ -0,0 +1,116 @@
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+import bs4 as bs
+import pandas as pd
+import re
+
+def download_url(url):
```
TheCleric, line 8: This looks really interesting, and would like to see this in a place like `datahelpers/http.py` so others could use it as well.

pybaseball/split_stats.py
```diff
@@ -0,0 +1,116 @@
...
+    if year is None: #provides scores from yesterday if date is not provided
+        url=f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year=Career&t={pitch_or_bat}"
+    else:
+        year=str(year)
+        url=f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year={year}&t={pitch_or_bat}"
+    html=download_url(url)
+    soup=bs.BeautifulSoup(html, 'lxml')
+    return soup
+
+def get_player_info(playerid=None, soup=None):
```
TheCleric, line 37: I think this would be a lot simpler if you used the `HTMLTableProcessor` class in `datasources/html_table_processor.py` along with xpath. However I also understand all the other BRef calls work this way right now.

TheCleric:
I'd also add that my preference **(which I am not trying to enforce, but I know would help push the project towards safer, more maintainable code)**, would be for this PR to include the following:

1. **Tests**. Integration tests at least (i.e., calling this function and checking that it returns results or errors for common usages and edge cases), unit tests (which would mock out the requests.get calls with data files so that they can be run without actually hitting BRef to validate only the logic here), or both.
    * This will also help PR reviewers have more confidence that this file does what it says it does.
    * My personal goal is to always have 100% test coverage (viewable by running `pytest tests --doctest-modules --cov=pybaseball --cov-report term-missing` from the project root, and looking at the coverage percentage of the files you have added, and adding tests to cover any lines showing as not being under test).
2. **MyPy typing hints** (e.g., `def download_url(url: str) -> bytes:`) so that we can run static analysis against the project to make sure that the inputs and outputs of this file match.
    * My personal goal is to always have 0 errors come back for the files I'm adding/modifying, along with any lines in other files that access those files (viewable by running `mypy .` from the project root, I'll usually use grep to filter it down since our project typing is a work in progress like so: `mypy . | grep split_stats`

schorrm:
We are not going to require unit tests from contributors at this juncture.

docs/split_stats.md
mwisnie5: Yea, an int will work and is probably the preferred type. I updated this to say Integer

pybaseball/split_stats.py
mwisnie5: Are you saying to create a new file at datahelpers/http.py? Or am I just not finding it?

pybaseball/split_stats.py
mwisnie5: I haven't tried using the HTMLTableProcessor class, though it looks pretty helpful for this sort of thing. The tricky part about this bbref page is that the table for the splits is wrapped in comments and the player info was all within <p> tags. I had to do an extra layer of identifying the comments and then parsing for the table to get to the result. Again, haven't tried using this class, but it seems unlikely to work because of that formatting.

TheCleric: Xpath is pretty powerful, and supports comments, so it should work. I'll play around with it and see.

schorrm:
Oops, shouldn't have done that as a direct commit. It may require you to pull this back to do further work.

pybaseball/split_stats.py
schorrm: How is this vs `requests.get`? Performance differences? Reasons to prefer it?

schorrm: I would put this in `utils.py`. Although I would like to hear out the benefits of this vs the unmodified `requests.get()`

pybaseball/split_stats.py
```diff
@@ -0,0 +1,116 @@
...
+    if pitching_splits is True:
+        pitch_or_bat='p'
+    else:
+        pitch_or_bat='b'
+    if year is None: #provides scores from yesterday if date is not provided
+        url=f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year=Career&t={pitch_or_bat}"
+    else:
+        year=str(year)
+        url=f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year={year}&t={pitch_or_bat}"
+    html=download_url(url)
```
schorrm, line 33: stuff like this -- can you run this through a formatter please?

schorrm:
So great PR, just a couple of things:
1. Can you please run this through an autoformatter?
2. Can you please add type annotations for the function calls?
3. Why the `download_url` and not `requests.get`?

schorrm:
Oh, and one more thing: can we be more clever with coercion to int vs. coercion to numeric? IIRC @TheCleric came up with a nice trick for this

pybaseball/split_stats.py
TheCleric: It is admittedly a LITTLE more complicated than I'd like, but this would replace most of `split_stats` and could be adapted for `get_player_info`.

I will preface this by saying, don't add this if you don't feel comfortable. I plan on refactoring a lot of the BRef code at some point, so this would just be another one on the list. 😆 

As well if you DO choose to use this, feel free to move the xpath addon functions to the `html_table_processor.py`

```python
import re
from typing import Any, List, TypeVar, Union

import lxml.etree
import pandas as pd
import requests

from pybaseball.datasources.html_table_processor import HTMLTableProcessor

_T = TypeVar('_T')
ListOrSelf = Union[_T, List[_T]]

@lxml.etree.FunctionNamespace(None)
def parse(context: Any, html_str: str) -> lxml.etree._Element:
    return lxml.etree.HTML(html_str).xpath('/html/body/*')[0]
    # return lxml.etree.fromstring(html_str)

@lxml.etree.FunctionNamespace(None)
def parse_comment(context: Any, html_comment: ListOrSelf[lxml.etree._Comment]) -> ListOrSelf[lxml.etree._Element]:
    if isinstance(html_comment, list):
        return [parse_comment(context, x) for x in html_comment]
    elif isinstance(html_comment, lxml.etree._Comment):
        html_str = lxml.etree.tostring(html_comment).decode('utf-8')
        html_str = html_str.replace("<!--", "").replace("-->", "")
        return parse(context, html_str)
    else:
        raise ValueError(f"Expected a Comment or List[Comment], got {type(html_comment)}")

@lxml.etree.FunctionNamespace(None)
def safe_text(context: Any, element: ListOrSelf[lxml.etree._Element], default: str = '') -> ListOrSelf[str]:
    if isinstance(element, list):
        return [safe_text(context, x, default) for x in element]
    elif isinstance(element, lxml.etree._Element):
        return element.text or default
    else:
        raise ValueError(f"Expected an Element or List[Element], got {type(element)}")

playerid = 'troutmi01'
pitch_or_bat = 'b'
year = 2019

bref_html = requests.get(f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year={year}&t={pitch_or_bat}").content.decode('utf-8')

bref_comment_tables = [
    # Why do you hurt me so lxml?
    lxml.etree.fromstring(lxml.etree.tostring(x)) for x in
    lxml.etree.HTML(bref_html).xpath('parse_comment((//*[contains(comment(), "overthrow table_container")])//comment())')
]

root_url = "https://www.baseball-reference.com"
headings_xpath = "({TABLE_XPATH}/thead//th)/descendant-or-self::*/text()"
data_rows_xpath = "({TABLE_XPATH}/tbody//tr)"
data_cell_xpath = "safe_text((td|th)/descendant-or-self::*)"

processor = HTMLTableProcessor(root_url, headings_xpath, data_rows_xpath, data_cell_xpath, None)

split_tables = [processor.get_tabular_data_from_element(x) for x in bref_comment_tables]
for index, table in enumerate(split_tables):
    table['Split Type'] = bref_comment_tables[index].xpath('//table/caption/text()')[0]

data = pd.concat(split_tables).reset_index(drop=True)
data['1B'] = data['H'] - data['2B'] - data['3B'] - data['HR']
data['Player ID'] = playerid
data = data.set_index(['Player ID', 'Split Type', 'Split'])
data = data.drop(index=['Split'], level=2)
data = data.dropna(axis=1, how='all')

print(data)
```

TheCleric:
> Oh, and one more thing: can we be more clever with coercion to int vs. coercion to numeric? @TheCleric and I came up with a nice trick for this in the FanGraphs code (snooping it off the data), can we apply this here without too much difficulty?

Yep, if we use the `html_table_processor` it will do it, but if we just want the type coercion, @mwisnie5 can use the `datahelpers.postprocessing.try_parse_dataframe` function:

```python
from .datahelpers import postprocessing

# data = Go get your data....

data = postprocessing.try_parse_dataframe(data)
```

This will attempt to coerce dates, ints, and floats from raw data.

pybaseball/split_stats.py
TheCleric: > How is this vs `requests.get`? Performance differences? Reasons to prefer it?

I'll let @wisnie5 comment on why they thought this would be useful, but the immediate utility I see are auto retries on failures.

mwisnie5: I actually built this function for another project, but it was to avoid errors when the website wasn't responding. I found that when I called sites a ton of times in a row, I would have requests get error responses. This function gives it a few more tries before failing and I am no longer seeing the no response errors. .

pybaseball/split_stats.py
mwisnie5: I would prefer to keep the original implementation on this. I like the consistency of what I have with the rest of the bbref parsing functions and honestly, I don't follow everything you have here.

pybaseball/split_stats.py
mwisnie5: Done

pybaseball/split_stats.py
TheCleric: Completely understand.

mwisnie5:
> ```python
> data = postprocessing.try_parse_dataframe(data)
> ```

mwisnie5:
> > Oh, and one more thing: can we be more clever with coercion to int vs. coercion to numeric? @TheCleric and I came up with a nice trick for this in the FanGraphs code (snooping it off the data), can we apply this here without too much difficulty?
> 
> Yep, if we use the `html_table_processor` it will do it, but if we just want the type coercion, @mwisnie5 can use the `datahelpers.postprocessing.try_parse_dataframe` function:
> 
> ```python
> from .datahelpers import postprocessing
> 
> # data = Go get your data....
> 
> data = postprocessing.try_parse_dataframe(data)
> ```
> 
> This will attempt to coerce dates, ints, and floats from raw data.

When I added in this postprocessing, the indexes for Splits and Split Types are also getting removed. I'm not sure why those are being dropped by the function. What benefits does this processing have over using to_numeric.

TheCleric:
> > > Oh, and one more thing: can we be more clever with coercion to int vs. coercion to numeric? @TheCleric and I came up with a nice trick for this in the FanGraphs code (snooping it off the data), can we apply this here without too much difficulty?
> > 
> > 
> > Yep, if we use the `html_table_processor` it will do it, but if we just want the type coercion, @mwisnie5 can use the `datahelpers.postprocessing.try_parse_dataframe` function:
> > ```python
> > from .datahelpers import postprocessing
> > 
> > # data = Go get your data....
> > 
> > data = postprocessing.try_parse_dataframe(data)
> > ```
> > 
> > 
> > This will attempt to coerce dates, ints, and floats from raw data.
> 
> When I added in this postprocessing, the indexes for Splits and Split Types are also getting removed. I'm not sure why those are being dropped by the function. What benefits does this processing have over using to_numeric.

Good point. It does nuke indexes. I should definitely fix that. The advantage it has is it tries to be more selective about its type coercion (dates vs ints vs floats vs percentages), whereas pd.to_numeric just makes everything it can a float. I did see some issues with getting `try_parse_dataframe` working locally with your code, but was successful with replacing your coerce with this snippet (which honestly I may just replace `try_parse_dataframe` with, as it's cleaner):

```python
for col in data.columns:
    for row in range(len(data)):
        data.iloc[row][col] = postprocessing.try_parse(data.iloc[row][col], col)
```

pybaseball/split_stats.py
```diff
@@ -0,0 +1,125 @@
...
+    session = requests.Session()
+    retry = Retry(connect=3, backoff_factor=0.5)
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount('http://', adapter)
+    session.mount('https://', adapter)
+    resp = session.get(url)
+    return resp.content
+
+
+def get_split_soup(playerid, year=None, pitching_splits=False):
```
schorrm, line 22: Can you please put type annotations on the functions?

pybaseball/split_stats.py
```diff
@@ -0,0 +1,125 @@
...
+        ptags = info[i].find_all('p')
+        fv = []
+        for j in range(len(ptags)):
+            InfoRegex = re.compile(r'>(.*?)<', re.DOTALL)
+            r = InfoRegex.findall(str(ptags[j]))
+            for k in range(len(r)):
+                pattern = re.compile('[\W_]+')
+                strings = pattern.sub(' ', r[k])
+                if strings and strings != ' ':
+                    fv.append(strings)
```
schorrm, line 60: The last thing to cover before this is ready, I think:
Can you provide a comment for this one block of code? Everything else is fairly self explanatory, this is the one thing I don't really understand here.

pybaseball/split_stats.py
mwisnie5: done

pybaseball/split_stats.py
mwisnie5: done

mwisnie5:
One other issue coming up now, with the change in the postprocessing to coerce, I was able to run and get the splits successfully, but now when I am putting it back into my broader project, I'm generating a new error from trying to aggregate non-numeric types. Give me a bit of time to sort that out before we finish this merge.

schorrm:
I fixed a couple bugs in project usage, and with the docs. it works for me now. I'm ready to merge this in when you are.

pybaseball/split_stats.py
```diff
@@ -0,0 +1,129 @@
...
+    session = requests.Session()
+    retry = Retry(connect=3, backoff_factor=0.5)
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount('http://', adapter)
+    session.mount('https://', adapter)
+    resp = session.get(url)
+    return resp.content
+
+
+def get_split_soup(playerid: str, year: int = None, pitching_splits: bool = False) -> bs.BeautifulSoup:
```
TheCleric, line 23: Technically an int can't be None.

To do this correctly, you have to include:

`from typing import Optional`

And then change this from `int` to `Optional[int]`

pybaseball/split_stats.py
```diff
@@ -0,0 +1,129 @@
...
+        url = f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year=Career&t={pitch_or_bat}"
+    else:
+        year = str(year)
+        url = f"https://www.baseball-reference.com/players/split.fcgi?id={playerid}&year={year}&t={pitch_or_bat}"
+    html = download_url(url)
+    soup = bs.BeautifulSoup(html, 'lxml')
+    return soup
+
+
+def get_player_info(playerid: str = None, soup=None) -> dict:
```
TheCleric, line 38: To do these make sure you have `from typing import Dict, Optional` at the top:

```suggestion
def get_player_info(playerid: Optional[str] = None, soup: bs.BeautifulSoup = None) -> Dict:
```

TheCleric: Also you say playerID can be None, but get_split_soup does not. If it's really optional, please make them agree. If it's not, please remove the None default (and don't make it `Optional[str]`, leave it as `str`)

pybaseball/split_stats.py
```diff
@@ -0,0 +1,129 @@
...
+        'Bats': fv[3],
+        'Throws': fv[5],
+        'Height': int(fv[6].split(' ')[0])*12+int(fv[6].split(' ')[1]),
+        'Weight': int(fv[7][0:3]),
+        'Team': fv[10]
+    }
+    return player_info_data
+
+
+def get_splits(playerid: str, year: int = None, player_info: bool = False, pitching_splits: bool = False) -> pd.DataFrame:
```
TheCleric, line 73: To do this make sure you have `from typing import Optional` at the top

```suggestion
def get_splits(playerid: str, year: Optional[int] = None, player_info: bool = False, pitching_splits: bool = False) -> pd.DataFrame:
```

mwisnie5:
> > > > Oh, and one more thing: can we be more clever with coercion to int vs. coercion to numeric? @TheCleric and I came up with a nice trick for this in the FanGraphs code (snooping it off the data), can we apply this here without too much difficulty?
> > > 
> > > 
> > > Yep, if we use the `html_table_processor` it will do it, but if we just want the type coercion, @mwisnie5 can use the `datahelpers.postprocessing.try_parse_dataframe` function:
> > > ```python
> > > from .datahelpers import postprocessing
> > > 
> > > # data = Go get your data....
> > > 
> > > data = postprocessing.try_parse_dataframe(data)
> > > ```
> > > 
> > > 
> > > This will attempt to coerce dates, ints, and floats from raw data.
> > 
> > 
> > When I added in this postprocessing, the indexes for Splits and Split Types are also getting removed. I'm not sure why those are being dropped by the function. What benefits does this processing have over using to_numeric.
> 
> Good point. It does nuke indexes. I should definitely fix that. The advantage it has is it tries to be more selective about its type coercion (dates vs ints vs floats vs percentages), whereas pd.to_numeric just makes everything it can a float. I did see some issues with getting `try_parse_dataframe` working locally with your code, but was successful with replacing your coerce with this snippet (which honestly I may just replace `try_parse_dataframe` with, as it's cleaner):
> 
> ```python
> for col in data.columns:
>     for row in range(len(data)):
>         data.iloc[row][col] = postprocessing.try_parse(data.iloc[row][col], col)
> ```

I'm going to revert this change back out. I found that by going through and editing each value in the df, its taking roughly triple the time to execute. This is really problematic for me since I am getting the splits for every player in the starting lineups in a game. Combined, this change made the process of getting the splits take 77.1 seconds compared to 24.4 seconds with the original implementation.

mwisnie5:
> I fixed a couple bugs in project usage, and with the docs. it works for me now. I'm ready to merge this in when you are.

I'm good with this now if you guys are ok with the changes I made to go back to my original implementation for coercing to numeric for the performance differences I mentioned above.

schorrm:
I want to put in a more performant thing for that.
Can you please try the following:
```python3
data = data.apply(pd.to_numeric, errors='coerce').convert_dtypes()
```
Should work just fine, and I think the perf hit should not be that bad.

pybaseball/split_stats.py
```diff
@@ -0,0 +1,127 @@
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+from datahelpers import postprocessing
```
schorrm, line 4: This needs to be `.datahelpers`, although it can also be removed I think

schorrm:
@mwisnie5 perf?

pybaseball/split_stats.py
mwisnie5: This was removed. I'm good with the performance and realized that the issue I had earlier was on how I am handling this df after the split stats function. I'm good if you are.

schorrm:
LGTM.

schorrm:
Actually -- what's going with the pitching stats? Not seeing those in there, only the batting stats in the pitching splits.

mwisnie5:
> Actually -- what's going with the pitching stats? Not seeing those in there, only the batting stats in the pitching splits.

What are you seeing? I'm still getting pitching splits when I run it with `pitching_splits=True`

mwisnie5:
@schorrm good catch on the missing pitching splits. The game level tables were getting dropped. Since the table headers are different between the two types of tables, I created two dataframes and concatted them at the end. I was a bit torn between concatting them or returning two dataframes. I ended up concatting them because even though it leaves NAs for big chunks of data where the two tables don't overlap, it seemed cleaner than returning two dataframes. I can change this if needed, but personally prefer this approach.

schorrm:
I think two dataframes would be better (either a tuple or a namedtuple).

TheCleric:
> I think two dataframes would be better (returned as either a tuple or a namedtuple).

Or to be consistent with a lot of our other stats, would it make sense to have two functions: `get_batting_splits` and `get_pitching_splits`?

schorrm:
So the tricky part is that there's batting splits, pitching batting splits, and pitching pitching splits, essentially. @mwisnie5 am I understanding correctly?

mwisnie5:
> So the tricky part is that there's batting splits, pitching batting splits, and pitching pitching splits, essentially. @mwisnie5 am I understanding correctly?

yea, for pitching splits there are the batting splits against the pitcher and then actual pitching splits.

mwisnie5:
Returned both pitching dataframes and removed ht, weight, and team from player info. Blame Pablo Sandoval (or bref) for that one.

pybaseball/split_stats.py
```diff
@@ -156,9 +156,9 @@ def get_splits(playerid: str, year: Optional[int] = None, player_info: bool = Fa
...
         level_data = level_data.loc[playerid]
-        data = pd.concat([data, level_data])
+        # data = pd.concat([data, level_data])
     if player_info == False:
-        return data
+        return data, level_data if 'level_data' in locals() else None
     else:
         player_info_data = get_player_info(playerid=playerid, soup=soup)
-        return data, player_info_data
+        return data, player_info_data, level_data if 'level_data' in locals() else None
```
schorrm: Okay, can we do this without using symbol table introspection and that ternary there?
just
```python
if condition:
    return data, player_info_data
return data
```

mwisnie5: Updated.

Issue #85, opened 2020-08-29
Suggested improvements


TheCleric:
I've been using this library for a little bit (and just submitted my first PR!) and have really been enjoying it.

I have a few suggestions that I'd like to make. I'm more than willing to do a lot of the work on these suggestions, but wanted to make sure I would be on the same page with other devs before I just start throwing code over the wall.

1. Add some unit/integration tests. This library would be ripe for adding some unit and integration tests. Both to help make sure code changes are good, but also as a way to determine if an external data source format has changed.

2. Refactor/reuse code. It seems that a lot of the code that is used for the same site (e.g., FanGraphs, BRef) is very repetitive. I think it would be great to distill some of this down to some shared code to help minimize issues when changes are made. Would do this after the above suggestion to ensure that the library is 100% backwards compatible.

3. Adding a caching layer. Sometimes when making calls to FanGraphs, the site will start rejecting your requests for exceeding their rate limit. Locally I wrote a caching wrapper that will save my DataFrames to CSV and return those if available. Integrating this into the library would allow others to take advantage.

4. Add an extra field to the FanGraphs results that returns a column that is joinable to the other data. E.g., team data or player data. Right now the best we'd be able to get is the FanGraphs integer id. We may have to set up our own internal mapping for teams. There are some other sources we could tap into, such as one of these for players:

> http://crunchtimebaseball.com/baseball_map.html
> https://www.smartfantasybaseball.com/tools/

Let me know what you folks think. Not trying to rock the boat, just trying to take this thing to the next level.

schorrm:
Re 4 -- we have mappings for the FG integer ID. You can join through those also if you want.
Re 1: yeah, that's a good idea. It would take time to roll out for all the prior stuff though.
2: Sure. There's a bunch of the stuff that is already good with reuse, there's plenty more that could be more precise.
3: Honestly, I think this is probably the _single_ most important feature thingy here. Perhaps should be #1 on our priorities in general. I would suggest that we open a separate issue to spec out how we want caching to work.

TheCleric:
Great, I'll break out an issue for 3, add some research and some ideas, and see what the community thinks.

For 1 and 2, what I was thinking is writing some FG and BREF shared code, and then slowly add unit tests and cutover per file, sending a small PR per.

For 4: Looking at the code, I think I see the player ID mapping. Let me know if you would mind me adding something similar for team IDs.

Thanks for the support and feedback. Looking forward to collaborate on this!

[...]

PR #138, opened 2020-09-29
Add team ids and player ids; team id mapper for Fangraphs

TheCleric:
Team ID Lookup - Fangraphs

`fangraphs_teams(season: int, league: str) -> pd.DataFrame`

Return a mapping dataframe to map Fangraphs' team id to Lahman data. This can then be used to map to BRef and Retrosheet data with Lahman's `teamIDBR` and `teamIDretro` columns.


TheCleric:
This (along with #128) will resolve the lingering issues I originally brought up in #85

schorrm:
I feel like I'm missing something here

TheCleric:
Right now we don't have an easy way to join Fangraphs data to other data sources for two reasons:
1) We don't actually return the Fangraphs id (either player or team) in any of our results
2) We had no way previously to join Fangraphs team data to other team data.

This PR solves both of those problems.

schorrm:
We had the team name, but I hear that. I thought FGID was in the player frames?

schorrm:
Is TeamIDFG just a number? Is it a constant one (I imagine it is)?
Can we just have a map of name:number and apply those?

TheCleric:
> Is TeamIDFG just a number? Is it a constant one (I imagine it is)?
> Can we just have a map of name:number and apply those?

I honestly thought about that. Actually I was thinking about storing the result of this (possibly along with the existing `first_season_map` in `utils.py`) as a CSV file that gets loaded (like `mlbstadiums.csv`) instead of hitting the actual API since this data will never change unless:

- A team gets added to the MLB
- A team gets dropped by the MLB
- A team moves cities

Let me know if you'd like to go to a local cache of this data instead. We can keep the function to regenerate if need be, but only return the cache for end users.

schorrm:
I _think_ this data should be *so* non-volatile, it might even make sense to just straight up hardcode it in (not even the cache) and then change in a bump if there's an expansion team or a rename (I think we might expect 1-2 changes / **decade**). AFAIK, the only change we would've had to make in the 2010s was LA Angels of Anaheim to LA Angels in 2016.

TheCleric:
> I _think_ this data should be _so_ non-volatile, it might even make sense to just straight up hardcode it in (not even the cache) and then change in a bump if there's an expansion team or a rename (I think we might expect 1-2 changes / **decade**). AFAIK, the only change we would've had to make in the 2010s was LA Angels of Anaheim to LA Angels in 2016.

To be clear, when I say a cache, I don't mean our new caching mechanism, I just mean we commit the data in the project as a CSV to keep us from having 100-200 lines of a dictionary definition.

schorrm:
That sounds correct to me

TheCleric:
FYI Tests are failing here due to a bug that is fixed in PR #144

schorrm:
LGTM

Issue #119, opened 2020-09-11
Non-issue/non-PR communication


TheCleric:
I've been submitting a lot of PRs lately (too many?) and I feel like some of them could have been cleared up with me asking some questions like "Would anyone mind if...?" or "Does anyone think this is a good idea?".

I don't really want to clog up our issues with these sorts of questions, so I was wondering what kind of interest other devs would have in doing some non-Github communications for quick questions/asides. Anything important could be brought into Github, but would keep this repo clean from unimportant discussions.

I initially thought email, but that can get clunky. I'm thinking IRC chat would work, and have added a channel on Freenode (an open source IRC server, also where python and linux have their IRC channels).

If that's okay here it is:
https://webchat.freenode.net/#pybaseball

Respond to this if anyone has alternative suggestions or if anyone thinks this is a bad idea.

schorrm:
I don't do IRC, so it doesn't strike me as great. I'm happy to have a discussion group, I'd just prefer it on a medium I might actually see.

TheCleric:
> I don't do IRC, so it doesn't strike me as great. I'm happy to have a discussion group, I'd just prefer it on a medium I might actually see.

Completely understand. What would be preferable for you?

schorrm:
Email, WhatsApp, FBM, whatever, just something that I already use

schorrm:
My main message services are FB and WhatsApp. Ppl can also PM me on Twitter if you like.

TheCleric:
I don't hand out my social media handles publicly, but if someone wants to email me: adamweeden@gmail.com, and I can send you them via that.

kmedved:
A few other packages I've seen use Slack for this sort of thing (e.g. Ray and Tabnet both have useful slack discussions, for issues which don't rise to the level of a Github issue). Not sure if you guys Slack, but wanted to throw that out there.

TheCleric:
> A few other packages I've seen use Slack for this sort of thing (e.g. Ray and Tabnet both have useful slack discussions, for issues which don't rise to the level of a Github issue). Not sure if you guys Slack, but wanted to throw that out there.

Yeah that would work for me as well as I'm on Slack all day for work anyway. 😆 

Discord would also work for me, but I know some folks in the open source world really don't like it as it's closed source and they've been known to monitor communication.

bdilday:
yeah, slack seems ideal

kmedved:
Obviously only makes sense if it's something @schorrm uses however.

[...]

PR #141, opened 2020-10-01
Add flag_imputed_data function

tjburch:
I've added a function to utils.py that will add a flag to any Statcast DataFrame with a boolean flag if it's possibly imputed as a result of the no-nulls approach in StatCast in the TrackMan era. For info on the no-nulls approach see [here](https://tht.fangraphs.com/43416-2/). This boolean can then be used for filtering if desired, the following pictures are launch speed and launch angle before and after filtering on this flag.

Before filtering

After filtering

The requirements have 3 criteria, specific combinations of launch angle, speed and bb type. If and only if all three criteria are simultaneously _exactly_ matched, the boolean is flipped to true.


schorrm:
This is _really_ good. Thank you very much!

pybaseball/utils.py
```diff
@@ -110,3 +111,58 @@ def get_text_file(url):
...
+def flag_imputed_data(statcast_df):
+    """Function to flag possibly imputed data as a result of no-nulls approach (see: https://tht.fangraphs.com/43416-2/)
+       Note that this imputation only occured with TrackMan, not present in Hawk-Eye data (beyond 2020)
+    Args:
+        statcast_df (pd.DataFrame): Dataframe loaded via statcast.py, statcast_batter.py, or statcast_pitcher.py
+    Returns:
+        pd.DataFrame: Copy of original dataframe with "possible_imputation" flag
+    """
+
+    ParameterSet = namedtuple('ParameterSet',"ev angle bb_type")
```
schorrm, line 124: Can this be changed to: `['ev', 'angle', 'bb_type']`? A bit clearer I think.

pybaseball/utils.py
```diff
@@ -110,3 +111,58 @@ def get_text_file(url):
...
+    impute_combinations.append(ParameterSet(ev=90.2, angle=-13.0, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=90.3, angle=-17.3, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=84.0, angle=-13.0, bb_type="ground_ball"))
+
+
+    df_return = statcast_df
+    df_return["possible_imputation"] = False
+    for param_set in impute_combinations:
+        bool_logic = (df_return["launch_speed"] == param_set.ev) & (df_return["launch_angle"] == param_set.angle) & (df_return["bb_type"] == param_set.bb_type)
+        df_return["possible_imputation"] = df_return["possible_imputation"] | bool_logic
```
schorrm, line 165: This is essentially equivalent to a join on the LS / LA / BBType, just without joining, only whether it would be joined.
Would making the params into a DataFrame with a column `imputed` and a left join work?
Is that a better or worse way to do this?

schorrm:
Also, can I please encourage you to make a PR with that visualization?

pybaseball/utils.py
tjburch: I think a left join should be equivalent. In my pandas experience, joins are more efficient than `for` loops, so probably the better approach of the two, but I'll give it a test today to check that's the case.

tjburch:
Sure, not a problem. Were you thinking just a short notebook in the `EXAMPLES` directory? Or something else?

tjburch:
The new push has the changes as indicated. As far as the merging via join, performance-wise the two ended up being basically equivalent, but it's a bit more readable. This screenshot shows timing benchmarks between the original code (temporarily renamed `flag_imputed_data_original` for the test) and via join (`flag_imputed_data_join`). In the commit, the name is the same as before.

Screen Shot 2020-10-02 at 7 40 07 AM

tjburch:
Ah - there's definitely a bug in that implementation. Don't merge just yet...

tjburch:
Resolved.

pybaseball/utils.py
```diff
@@ -110,3 +111,61 @@ def get_text_file(url):
...
+
+    # pop-ups
+    impute_combinations.append(ParameterSet(ev=80, angle=69, bb_type="popup"))
+    impute_combinations.append(ParameterSet(ev=37, angle=62, bb_type="popup"))
+    impute_combinations.append(ParameterSet(ev=86, angle=67, bb_type="popup"))
+
+    # Flyout
+    impute_combinations.append(ParameterSet(ev=71.4, angle=36.0, bb_type="fly_ball"))
+    impute_combinations.append(ParameterSet(ev=89, angle=39, bb_type="fly_ball"))
+    impute_combinations.append(ParameterSet(ev=89.2, angle=39.3, bb_type="fly_ball"))
```
bdilday, line 135: when I look at the statcast data, I see a lot of values at `ev=89.2`, `angle=39.0`. and not `ev=89.2`, `angle=`39.3`. I'm wondering if this record with `angle=39.3` might be a typo?

bdilday: in fact it looks like maybe all the imputed values have angle that are integers?

pybaseball/utils.py
```diff
@@ -110,3 +111,61 @@ def get_text_file(url):
...
+    impute_combinations.append(ParameterSet(ev=90, angle=-17, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=90.2, angle=-13.0, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=90.3, angle=-17.0, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=90.3, angle=-17.3, bb_type="ground_ball"))
+    impute_combinations.append(ParameterSet(ev=84.0, angle=-13.0, bb_type="ground_ball"))
+
+
+    df_imputations = pd.DataFrame(data=impute_combinations)
+    df_imputations["possible_imputation"] = True
+    df_return = statcast_df.merge(df_imputations, how="left",
```
bdilday, line 165: because of the merge here, `df_return` has columns `ev` and `angle`, which the original didn't have and which are either null or identical to `launch_speed`, `launch_angle`. It might be better to drop the `ev` and `angle` columns before returning, or just name them `launch_speed` and `launch_angle` in the `impute_combinations` data to begin with.

schorrm:
@tjburch maybe even a library function -- e.g. `pybaseball.plot_bb_profile()`?

tjburch:
Working through these proposed edits now, thanks for the eyes on the code.

tjburch:
> @tjburch maybe even a library function -- e.g. `pybaseball.plot_bb_profile()`?

Just committed a pretty generic mock-up of this. One question was if there were any preferences about mpl figure size, I set it as dpi=300 for now, but can edit it.

TheCleric:
FYI Tests are failing here due to a bug that is fixed in PR #144

schorrm:
That looks good, but you can drop it from this PR and submit separately?

schorrm:
I am thinking, by the ways, that it makes sense to expose this primarily as a boolean `flag_imputed_data=True` in the statcast batting calls, rather than as a function separately?

pybaseball/utils.py
tjburch: So that originally came from the tables found [here](https://tht.fangraphs.com/43416-2/), under "Stringer Fly Balls," then rounded to the precision in the DFs we get. I'll do some more investigating though to see if something was missed.

tjburch:
>@tjburch That looks good, but you can drop the visualization from this PR and submit separately?

Done

>I am thinking, by the ways, that it makes sense to expose this primarily to the user as a boolean flag_imputed_data=True in the statcast batting calls, rather than as a separate thingy?

Sure, we can do that. For what it's worth, the original implementation was loosely motivated from a similar function in [baseballr](https://github.com/BillPetti/baseballr/blob/master/R/label_statcast_imputed_data.R), but we're not held to their approach by any means. I'll get that implementation asap.

pybaseball/utils.py
bdilday: it looks like they've started rounding all angles to integer?

bdilday:
> Sure, we can do that. For what it's worth, the original implementation was loosely motivated from a similar function in [baseballr](https://github.com/BillPetti/baseballr/blob/master/R/label_statcast_imputed_data.R), but we're not held to their approach by any means. I'll get that implementation asap.

I'm the person that implemented the [`baseballr` version](https://github.com/BillPetti/baseballr/pull/71), BTW. I can say the reason it's a separate function instead of an argument to the statcast scrape function because it was considered "experimental" and I don't think `baseballr` core maintainer(s) wanted it in the core API at that time. Seems like it's no problem to include it as option in `pybaseball`.

BTW I opened a [PR at `baseballr` yesterday](https://github.com/BillPetti/baseballr/pull/177) that adds the script I used to derive the values,

pybaseball/utils.py
tjburch: Ah yeah - so I guess it varies by year... pains :( Maybe this is a product of HawkEye? As I commented below, probably your thought of using a heuristic derivation is the right move.

bdilday:
> @bdilday Thanks for that! In your code I see:
> 
> ```
> # use 5 here? some other number? 99.X percentile? this is why I referred to 
> # it as a heuristic in the `label_statcast_imputed_data` documentation
> ```
> 
> And the 5 value ultimately used. Do you find this heuristic works? I could also think if you did some percentage of total rows of bb_type (e.g. if there's 500 fly balls, probably no more than 5 should be a given EV/LA, but if there's 10,000 maybe that number should go up?)

It worked for the data set I was using at the time, but yeah it's not appropriate for general usage. I think what you mentioned, setting some threshold of percentage of overall balls, is a great starting point, and it could be refined in subsequent PRs if there's a need.

schorrm:
I will defer to your judgment on this one.

schorrm:
@tjburch @bdilday what's the status on this?

tjburch:
I have a notebook with a more motivated derivation of values - will submit a PR with it in the next hour so the values can be agreed on, then update this PR with those values.

tjburch:
Updated numbers in accordance with the derivation in PR #149

schorrm:
@tjburch is this ready to go?
Also, check out the dev slack -- see #119

tjburch:
Yep! Good to go.

PR #142, opened 2020-10-02
True unit testing refactor

TheCleric:
This PR only does the following:

- Adds some fixtures to prevent requests calls inside unit tests to ensure that we're doing unit tests (i.e., logic) not integration tests within our unit testing folder
    - This will help us to see where we have missing code coverage in unit testing to make sure we're completely testing all logical code paths
- Moves any unit tests that failed after the above was implemented to the integration tests folder.

NO UNIT TESTS WERE HARMED IN THE MAKING OF THIS PULL REQUEST, ONLY RELOCATED FOR THEIR SAFETY. 😄


tests/conftest.py
```diff
@@ -6,14 +6,15 @@
...
+
+from pybaseball import cache
 
 
 @pytest.fixture(name='logging_side_effect')
 def _logging_side_effect() -> Callable:
     def _logger(name: str, after: Optional[Callable] = None) -> Callable:
         def _side_effect(*args: Any, **kwargs: Any) -> Optional[Any]:
-            logging.debug(f'Mock {name} => {args} {kwargs}')
+            logging.debug('Mock %s => %s %s', name, args, kwargs)
```
schorrm, line 17: why?

TheCleric: Not my preferred way, but it's the Python preferred way for the logging library, and actually faster as it allows it to lazy load the debug string:

https://docs.python.org/3/howto/logging.html#optimization

schorrm: I think you're misreading that. The speed is from eliding it with the `loggingIsEnabled` guard. And in general, f-strings are faster than other methods in Python.

TheCleric: > I think you're misreading that.

I am and I'm not. 🤣 

I agree I totally breezed by the `loggingIsEnabled` section, so you're right there and that's on me. However, the reason I got there in the first place is because pylint complains about this line. Which led me to research why. Which led me to long threads like this one:

https://github.com/PyCQA/pylint/issues/2395

Where the arguments boil down to two reasons:
1. If the f-string throws an exception on creation, your logging fails
2. String interpolation costs SOMETHING (however minimal), and doing it this way means that the logging library will skip it on the backend if the log will be discarded (e.g., we're logging at WARNING, and this is a DEBUG), as opposed to us doing the string interpolation every time regardless of log level.

Honestly I'm not concerned with number 1.

For number 2, I could see the argument that the readability gain is worth the very minor efficiency gain.

I'll defer to your judgement as I could go either way on it.

schorrm: I'm not going to get onboard with overoptimizing the testing process. Stick to the f-string for readability

TheCleric: @schorrm I reverted it back, and added some configs so that pylint won't flag this again.

PR #143, opened 2020-10-03
Add tests for amateur_draft

TheCleric:


pybaseball/amateur_draft.py
```diff
@@ -3,33 +3,42 @@
...
     res = requests.get(url, timeout=None).content
     draft_results = pd.read_html(res)
     return draft_results
 
+
 @cache.df_cache()
-def amateur_draft(year, round, keep_stats=True):
-    draft_results = get_draft_results(year, round)
+def amateur_draft(year: int, draft_round: int, keep_stats: bool = True) -> pd.DataFrame:
+    draft_results = get_draft_results(year, draft_round)
```
schorrm, line 19: Okay. I have bad feelings about this, because on the one hand, this absolutely shouldn't have been allowed in the first place, OTOH it is an API change. I think change it, but change the docs to match.

TheCleric: We're on the EXACT same page regarding this. I didn't want to change an API parameter name, but when it's a reserved word...

Great point about the docs, I'll fix those.

TheCleric: @schorrm Docs updated.

PR #145, opened 2020-10-05
Add plot_bb_profile

tjburch:
Add plot_bb_profile.py as discussed in PR #141. Splits up a DataFrame by bb_type and will plot a passed StatCast parameter for each type. Defaults to `launch_angle`, just so it has something to go off of for a generic call, but can be generalized to any parameter.


tjburch:
As I was submitting this, I noticed that there does already seem to be a `plotting.py`, maybe this would be better to be a function in there? Thoughts?

tjburch:
Done.

TheCleric:
> Done.... appears to have an issue on the matplotlib dependency? Not sure what the action needed here is.

Probably need to add matplotlib to `install_requires` in `setup.py`.

tjburch:
That worked. Thanks @TheCleric!

tjburch:
>Is 3.3.0 critical here? Do you have any real clue which version it would need to be?

No, not critical. Nothing used in mpl is even relatively new - any suggestions of a version to use? Maybe just like the most recent major revision? e.g. 3.0.0?

TheCleric:
> > Is 3.3.0 critical here? Do you have any real clue which version it would need to be?
> 
> No, not critical. Nothing used in mpl is even relatively new - any suggestions of a version to use? Maybe just like the most recent major revision? e.g. 3.0.0?

3.0.0 would likely work. What I do in these situations is I'll try prior versions in the setup.py until it breaks, and then go to the oldest working version.

tjburch:
@TheCleric silly question, I don't have a ton of experience here. Is there a way to cleanly run the tests locally so I'm not doing a bunch of commits to run with github's CI?

TheCleric:
> @TheCleric silly question, I don't have a ton of experience here. Is there a way to cleanly run the tests locally so I'm not doing a bunch of commits to run with github's CI?

Not a silly question at all! To do so, just make sure you have the package installed with the test dependencies like so (from the project root):

`pip install -e .[test]`

Once that's done it's as simple as running `pytest`

schorrm:
Looks good. Last thing is we just need to update the docs accordingly.

tjburch:
Rolled back to appropriate dependency, think this should be okay now.

schorrm:
Great. Now we just need to update the plotting docs

schorrm:
LGTM, merging once tests are done

PR #146, opened 2020-10-05
Statcast testing/minor refactor

TheCleric:
This PR adds statcast testing.

After adding the testing, some minor refactors were done to make the `statcast.py` module more easily digestible, mostly in the way retries are handles in `statcast_large_request`, along with making sure pylint and mypy are happier here.


pybaseball/statcast.py
```diff
@@ -35,106 +98,53 @@ def large_request(start_dt_date: date, end_dt_date: date, step: int, verbose: bo
...
     print("This is a large query, it may take a moment to complete")
-    
-    # Number of days per mini-query
-    # (test this later to see how large I can make this without losing data)
 
     # While intermediate query end_dt <= global query end_dt, keep looping
-    d = d1 + timedelta(days=step)
-    while d <= d2: 
+    current_end_date = current_start_date + timedelta(days=step)
+    while current_end_date <= end_dt_date:
```
schorrm, line 110: I don't like this refactor. `end_dt_date` is an ugly and redundant name, and stuff is too long here in general.
I would suggest that we do:
`c`, and leave end as just `end_dt`
so `while subq_end <= end_dt: ...` etc
I'm not sure about the `current_end_date`, I am sure that `end_dt_date` should just be `end_dt`

TheCleric: `current_end_date` is an okay name, though long (I like your suggestion of `subq_end` though, I'll fix to that).

As far as the `end_dt_date` you're completely right. This got in because I refactored this slowly, and originally when this function took strings (like `end_dt`) AND dates, then `end_dt_date` helped distinguish it from the string value. But that's gone and this remained. I'll fix.

schorrm:
hmmm how I do redo this?

PR #148, opened 2020-10-05
Move existing CSV to new data folder + cleanup

TheCleric:
This PR started as a straight move of the mlbstadiums.csv file to the new data folder for consistency. However in doing so a few things were discovered:

- When testing the move to ensure nothing broke, I noticed that the second plotting example was incomplete. I worked backwards to fill it in to match the existing image.
- When doing that, I noticed the docstring for the `spraychart` function incorrectly specified how to use the `colorby` argument.
- While updating all references to the existing csv file, I added the fangraphs csv file to the `MANIFEST.in` for consistency. This likely does nothing as I believe `MANIFEST.in` files are only for python 2.x, but just to be safe...


pybaseball/plotting.py
```diff
@@ -50,7 +50,7 @@ def spraychart(data, team_stadium, title='', tooltips=[], size=100,
     tooltips: list of variables in data to include as tooltips
     size: size of marks on plot
     colorby: which category to color the mark with. Events or player name.
-           must be 'events' or 'name'
+           must be 'events' or 'player'
```
TheCleric, line 53: Check vs. Line 81 below

schorrm:
LGTM

PR #149, opened 2020-10-06
Add derivation notebook for imputed values

tjburch:
This PR adds a notebook where I did a quick study to see what values occur disproportionally often in StatCast data, following discussion in PR #141, where it seemed a motivated heuristic was preferable over copying values from the FG article. 

Basically, I looked at when a given EV/launch angle pair comprised more that 0.2% of the data for a given bb_type for a given year. About halfway through the notebook, there's a table showing the pairs, bb_type, and years which they pass the threshold. Those values are:

| launch_angle | launch_speed |     bb_type | Years above threshold |
|-------------:|-------------:|------------:|----------------------:|
|        -21.0 |         82.9 | ground_ball |                     5 |
|        -17.0 |         90.3 | ground_ball |                     5 |
|         15.0 |         90.4 |  line_drive |                     2 |
|         18.0 |         91.1 |  line_drive |                     3 |
|         30.0 |        102.8 |    fly_ball |                     1 |
|         39.0 |         89.2 |    fly_ball |                     5 |
|         61.0 |         79.9 |       popup |                     1 |
|         62.0 |         86.3 |       popup |                     1 |
|         69.0 |         80.0 |       popup |                     5 |
|         70.0 |         72.3 |       popup |                     1 |

I also show plots before/after removing all these values. I propose removing all of these, but perhaps we should open discussion here. Once settled, I'll update PR #141.

Tagging @bdilday here.


bdilday:
I think these look like the right values, but one comment about the notebook is it shows the before and after based on a given threshold, but doesn't give a way to choose a threshold. It'd be helpful to have an additional function that; groups the data by bb_type, events, year; computes the fractions; orders the data inversely by fraction; and displays the top 50 rows or whatever. This would help eyeball a good threshold value. Then this threshold could be applied in the subsequent part of the notebook that shows the before / after. Does that make sense?

tjburch:
Sure - I basically did that to motivate the 0.002; I can rewrite that code to make it transparent.

tjburch:
See updated notebook - printed out as @bdilday suggested. Updated proposal is to flag the following as possibly imputed:

| launch_angle | launch_speed |     bb_type | Years above threshold |
|-------------:|-------------:|------------:|----------------------:|
|        -21.0 |         82.9 | ground_ball |                     5 |
|        -17.0 |         90.3 | ground_ball |                     5 |
|         15.0 |         90.4 |  line_drive |                     2 |
|         18.0 |         91.1 |  line_drive |                     3 |
|         30.0 |        102.8 |    fly_ball |                     1 |
|         39.0 |         89.2 |    fly_ball |                     5 |
|         69.0 |         80.0 |       popup |                     5 |

bdilday:
LGTM! I think I'd move the file out of `pybaseball/pybaseball` to `pybaseball/EXAMPLES` but I defer to @schorrm on that

schorrm:
@bdilday seconded

tjburch:
Done

PR #150, opened 2020-10-06
Function to add spray angle to Statcast dataframes

tjburch:
This PR adds `pybaseball/datahelpers/statcast_utils.py`, a file to keep functions that operate on statcast dataframes. I chose to create a new file over adding to `datahelpers/transforms.py` since that seems to be oriented more toward Lahman dataframes.

In that file, I've added a function which calculates spray angle, as well as an adjusted spray angle which flips the angle sign for left-handed batters, in order to make it a push/pull angle. The angle added is toggled by a boolean argument `adjusted`, when false (default) it adds the raw spray angle, when true it adds the adjusted one.

In addition, a documentation file was added. I included a graphic here, which made me a little concerned about the number of images accruing in the docs directory, so I created a sub-directory named `images` to house any further graphics. This required a few adjustments in `plotting.md` to point to the moved images.


pybaseball/datahelpers/statcast_utils.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+import pandas as pd
+from typing import Optional
+
+def add_spray_angle(df: pd.DataFrame, adjusted: Optional[bool] = False):
```
TheCleric, line 5: Please specify return type:

```suggestion
def add_spray_angle(df: pd.DataFrame, adjusted: Optional[bool] = False) -> pd.DataFrame:
```

As well, `Optional` is only needed of `None` is a valid value to pass to `adjusted`. Otherwise you can just leave it as `bool`

tjburch: Thanks - I've not used type hinting a ton, and actually just learned I was misusing Optional a couple of hours ago :) resolved in the latest commit. Thanks for the check.

schorrm:
LGTM

PR #152, opened 2020-10-09
Statcast testing/minor refactor (Part Deux)

TheCleric:
This PR adds statcast testing.

After adding the testing, some minor refactors were done to make the statcast.py module more easily digestible, mostly in the way retries are handles in statcast_large_request, along with making sure pylint and mypy are happier here.


schorrm:
Note for the record: there was a glitch in the merge of #146 -- this is the replica

PR #157, opened 2020-10-15
Change how we postprocess statcast to improve performance

### LINKED ISSUE #154: Severe Perf issue in Statcast
When calling data = statcast('2020-08-01', '2020-10-05') on a large query it just runs forever now.

TheCleric:
Fixes #154

### DIFF
--- a/pybaseball/datahelpers/postprocessing.py
+++ b/pybaseball/datahelpers/postprocessing.py
@@ -18,24 +18,58 @@ date_formats = [
     (re.compile(r'^\d{4}-\d{1,2}-\d{1,2}T\d{2}:\d{2}:\d{2}.\d{1,6}Z$'), '%Y-%m-%dT%H:%M:%S.%fZ'),
 ]
 
+
 def try_parse_dataframe(
     data: pd.DataFrame,
     null_replacement: Union[str, int, float, datetime] = np.nan,
     known_percentages: List[str] = []
 ) -> pd.DataFrame:
-    values = [
-        {column: try_parse(data[column][row_i], column) for column in data.columns}
-        for row_i in range(len(data)) 
-    ]
+    data_copy = data.copy()
+    data_copy = coalesce_nulls(data_copy, null_replacement)
+    data_copy = data_copy.apply(pd.to_numeric, errors='ignore', downcast='integer').convert_dtypes(convert_string=False)
 
-    return pd.DataFrame(values)
+    rows = len(data_copy.values)
+
+    for column in data_copy.columns:
+        if str(data_copy[column].dtype) in ['object', 'string']:
+            # Only check the first value of the column and test that;
+            # this is faster than blindly trying to convert entire columns
+            first_value_index = data_copy[column].first_valid_index()
+            if first_value_index is None:
+                # All nulls
+                continue
+            first_value = data_copy[column].loc[first_value_index]
+
+            if str(first_value).endswith('%') or column.endswith('%') or column in known_percentages:
+                column_pos = data_copy.columns.get_loc(column)
+                for row in range(rows):
+                    data_copy.iat[row, column_pos] = try_parse(
+                        data_copy.iat[row, column_pos],
+                        column,
+                        null_replacement=null_replacement,
+                        known_percentages=known_percentages
+                    )
+            else:
+                # Doing it this way as just applying pd.to_datetime on
+                #the whole dataframe just tries to gobble up ints/floats as timestamps
+                for date_regex, date_format in date_formats:
+                    if isinstance(first_value, str) and date_regex.match(first_value):
+                        data_copy[column] = data_copy[column].apply(pd.to_datetime, errors='ignore', format=date_format)
+                        data_copy[column] = data_copy[column].convert_dtypes(convert_string=False)
+                        break
+
+    return data_copy.convert_dtypes(convert_string=False)
+
 
 def try_parse(
-    value: str,
+    value: Union[None, str, int, datetime, float],
     column_name: str,
     null_replacement: Union[str, int, float, datetime] = np.nan,
     known_percentages: List[str] = []
 ) -> Union[str, int, float, datetime]:
+    if value is None:
+        return null_replacement
+
     if not isinstance(value, str):
         return value
 
@@ -65,10 +99,10 @@ def try_parse(
     return value
 
 
-def coalesce_nulls(data: pd.DataFrame, value: Union[str, int, float] = np.nan) -> pd.DataFrame:
+def coalesce_nulls(data: pd.DataFrame, value: Union[str, int, float, datetime] = np.nan) -> pd.DataFrame:
     # Fill missing values with NaN
     for regex in null_regexes:
-        data.replace(regex, value, regex=True, inplace=True)
+        data.replace(regex.pattern, value, regex=True, inplace=True)
 
     return data
 
@@ -77,7 +111,7 @@ def columns_except(data: pd.DataFrame, columns: List[str]) -> List[str]:
     return list(np.setdiff1d(data.columns, columns))
 
 
-def convert_numeric(data: pd.DataFrame, numeric_columns: List[str]):
+def convert_numeric(data: pd.DataFrame, numeric_columns: List[str]) -> pd.DataFrame:
     # data.loc[data[numeric_cols] == ''] = None
     # data[numeric_cols] = data[numeric_cols].astype(float)
     # Ideally we'd do it the pandas way ^, but it's barfing when some columns have no data
@@ -150,8 +184,10 @@ def augment_lahman_pitching(stats_df: pd.DataFrame) -> pd.DataFrame:
     """
     return stats_df
 
+
 def aggregate_by_season(stats_df: pd.DataFrame) -> pd.DataFrame:
     return stats_df.groupby(["playerID", "yearID"]).sum().reset_index()
+
 
 def check_is_zero_one(instance, attribute, value):
     if value not in [0, 1]:
--- a/pybaseball/datahelpers/postprocessing.py
+++ b/pybaseball/datahelpers/postprocessing.py
@@ -1,6 +1,6 @@
 import re
 from datetime import datetime
-from typing import List, Union
+from typing import List, Optional, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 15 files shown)


bdilday:
here's a thought - do we need to automatically detect data type? after all `pybaseball` doesn't ingest arbitrary data, it ingests data from a limited set of known sources, and we know the schema for each table it ingests. could we define a schema(s) instead?

TheCleric:
> here's a thought - do we need to automatically detect data type? after all `pybaseball` doesn't ingest arbitrary data, it ingests data from a limited set of known sources, and we know the schema for each table it ingests. could we define a schema(s) instead?

I thought about that as well. I think that would be a good next step.

schorrm:
@bdilday - I like that -- I think if we fail this we should drop back to the prior (2.0) behavior, and throw a warning that our schema is out of date, but this can be the approach

pybaseball/datahelpers/postprocessing.py
```diff
@@ -18,24 +18,58 @@
...
-        for row_i in range(len(data)) 
-    ]
+    data_copy = data.copy()
+    data_copy = coalesce_nulls(data_copy, null_replacement)
+    data_copy = data_copy.apply(pd.to_numeric, errors='ignore', downcast='integer').convert_dtypes(convert_string=False)
+
+    rows = len(data_copy.values)
+
+    for column in data_copy.columns:
+        if str(data_copy[column].dtype) in ['object', 'string']:
```
bdilday, line 34: I think you can avoid a level of nesting here if you pre-filter the columns, e.g., 
```
string_columns = [dtype_tuple[0] for dtype_tuple in data_copy.dtypes.items() if str(dtype_tuple[1]) in ["object", "string"]]
for column in string_columns:
  first_value_index = data_copy[column].first_valid_index()
  .
  . etc
```

pybaseball/datahelpers/postprocessing.py
```diff
@@ -18,24 +18,58 @@
...
+            first_value_index = data_copy[column].first_valid_index()
+            if first_value_index is None:
+                # All nulls
+                continue
+            first_value = data_copy[column].loc[first_value_index]
+
+            if str(first_value).endswith('%') or column.endswith('%') or column in known_percentages:
+                column_pos = data_copy.columns.get_loc(column)
+                for row in range(rows):
+                    data_copy.iat[row, column_pos] = try_parse(
```
bdilday, line 46: I don't think you need to use the `try_parse` function anymore, since you know you're trying to parse percentages. It's also redundant to pass `column` to `try_parse` when you've already checked the column. Maybe we need a new function here, `parse_percentage` that's more limited in scope? 

even better would be to use pandas vectorized operations to avoid the explicit loop over rows. maybe something like, 
```
data_copy[column] = data_copy[column].str.replace("%", " ").astype(float) / 100
```
?

pybaseball/datahelpers/postprocessing.py
```diff
@@ -18,24 +18,58 @@
...
     null_replacement: Union[str, int, float, datetime] = np.nan,
     known_percentages: List[str] = []
 ) -> pd.DataFrame:
-    values = [
-        {column: try_parse(data[column][row_i], column) for column in data.columns}
-        for row_i in range(len(data)) 
-    ]
+    data_copy = data.copy()
+    data_copy = coalesce_nulls(data_copy, null_replacement)
+    data_copy = data_copy.apply(pd.to_numeric, errors='ignore', downcast='integer').convert_dtypes(convert_string=False)
```
bdilday, line 29: you're using `convert_types` twice, once here and once just before returning from the function, can we get the same overall effect with just applying `convert_types` just once?

schorrm, line 29: Is downcast correct here? Can you explain the particular chain here?

TheCleric: `downcast='integer'` should try to cast it to an integer if possible, but will settle for a float if that's what the data is.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html

> If not None, and if the data has been successfully cast to a numerical dtype (or if the data was numeric to begin with), downcast that resulting data to the smallest numerical dtype possible according to the following rules:
> 
> ‘integer’ or ‘signed’: smallest signed int dtype (min.: np.int8)
> 
> ‘unsigned’: smallest unsigned int dtype (min.: np.uint8)
> 
> ‘float’: smallest float dtype (min.: np.float32)
> 
> As this behaviour is separate from the core conversion to numeric values, any errors raised during the downcasting will be surfaced regardless of the value of the ‘errors’ input.
> 
> In addition, downcasting will only occur if the size of the resulting data’s dtype is strictly larger than the dtype it is to be cast to, so if none of the dtypes checked satisfy that specification, no downcasting will be performed on the data.

schorrm: Is an int8 good though? A single season's G should be an int8, a career won't be though

TheCleric: It just tries to go to the smallest int type, if the data is larger than that it will use an appropriate int type. For example I tested it on '4294967296' (2^32) and got the correct int back:

```python
            raw_data = pd.DataFrame(
            [
                ['1', 'TBR', '2019-01-01', '0.5', '40%', 8, '1048576'],
                ['2', 'NYY', '2019-02-01', '0.6', '20%', 'null', '4294967296']
            ],
            columns=['id', 'team', 'dt', 'rate', 'chance', 'wins', 'long_number']
        )
        processed_data = postprocessing.try_parse_dataframe(raw_data)
        print(processed_data)
```

```
    id team  dt           rate    chance  wins  long_number
0   1  TBR   2019-01-01   0.5     0.4     8     1048576
1   2  NYY   2019-02-01   0.6     0.2     <NA>  4294967296
```

schorrm: That wasn't my concern. My concern is does it automatically upcast instead of overflowing?

schorrm: which it does

bdilday: @schorrm in your comment above, "it does" means it upcasts or it overflows?

seems safer to just use `int64`; example where it could be a problem, 
```
>>> df1 = pd.DataFrame({'G': [127]})
>>> df2 = pd.DataFrame({'G': [127]})
>>> df1 = df1.apply(pd.to_numeric, errors='ignore', downcast='integer')
>>> df2 = df2.apply(pd.to_numeric, errors='ignore', downcast='integer')
>>> df1.G+df2.G
0   -2
```

schorrm: Good catch. I tried sum, I didn't get this one. No downcasting then

TheCleric:
I think I've addressed everything being asked for here. Let me know if I missed anything.

pybaseball/datahelpers/postprocessing.py
TheCleric: Well if that's the case we're back to some sort of schema type solution to say which columns should be which types.

pybaseball/datahelpers/postprocessing.py
```diff
@@ -54,21 +81,30 @@ def try_parse(
...
 
     return value
 
 
-def coalesce_nulls(data: pd.DataFrame, value: Union[str, int, float] = np.nan) -> pd.DataFrame:
+def try_parse_percentage(value: str) -> float:
+    if '.' in value:
+        return float(value.strip(' %')) / 100.0
+    else:
+        result = int(value.strip(' %'))
```
bdilday, line 100: I think if you cast as int and then divide by 100.0 you'll change it to float, so that this is exactly the same as `float(value.strip(' %')) / 100.0` from above.

pybaseball/datahelpers/postprocessing.py
```diff
@@ -18,24 +18,51 @@
...
 ) -> pd.DataFrame:
-    values = [
-        {column: try_parse(data[column][row_i], column) for column in data.columns}
-        for row_i in range(len(data)) 
-    ]
+    data_copy = data.copy()
+    data_copy = coalesce_nulls(data_copy, null_replacement)
+    data_copy = data_copy.apply(pd.to_numeric, errors='ignore', downcast='integer').convert_dtypes(convert_string=False)
+
+    rows = len(data_copy.values)
```
bdilday, line 31: rows doesn't seem to be used

pybaseball/datahelpers/postprocessing.py
schorrm: Well we're never gonna overflow a 64. Convert to int as possible, but not i8 I guess?

TheCleric: I don't think you can choose. If you say to "integer" (or "signed or "unsigned") it uses the smallest possible one it can. Only other option is float.

After you pointed this out I did an experiment where as long as you use pandas' sum function it will upcast it:

```python
>>> import pandas as pd
>>> df = pd.DataFrame({'G': ['127', '127', '127']})
>>> df = df.apply(pd.to_numeric, errors='ignore', downcast='integer')
>>> df['G'].sum()
381
```

I'll also play around with Pandera and see how it may help.

schorrm: Why do you need to downcast at all though?

TheCleric: This is the only way in pd.to_numeric to get it to coerce integers. Otherwise you end up with a bunch of XXX.0 floats for int columns.

schorrm: Why aren't we using the `convert_dtypes` here?

TheCleric: We are.

schorrm: @TheCleric this is what I'm talking about, why do we need `to_numeric` and `downcast` and `convert_dtypes`?

TheCleric: @schorrm I think I see what you are saying now. I think I have a leaner solution incoming now.

TheCleric:
For the record I created a branch locally to replace this with Pandera. Good news is I got it to work, bad news is it's around 2x slower. :(

schorrm:
omg gretchen, stop trying to make pandera happen, it's not gonna happen

pybaseball/datahelpers/postprocessing.py
TheCleric: Mostly because I was trying to keep this function generic for others to use. We can eliminate the numeric parsing or make it optional for statcast to bypass.

TheCleric:
> omg gretchen, stop trying to make pandera happen, it's not gonna happen

🤣 🤣 🤣 🤣 🤣 

Just experimenting. :)

TheCleric:
I made numeric parsing optional in this function since statcast's data comes in as numeric.

pybaseball/datahelpers/postprocessing.py
schorrm: can we eliminate the downcast? should be unnecessary because `convert_dtypes` should take care of making things ints?

schorrm:
We still want the dtype resolution though -- to int, etc. That was the whole point of this originally

TheCleric:
> We still want the dtype resolution though -- to int, etc. That was the whole point of this originally

That still works. The ints are coming in as ints. I think in 2.0.0 the reason it did not work is because we were casting all numeric_cols to float.

jldbc:
What's the final time measurement for this vs. before all the postprocessing changes?

TheCleric:
> What's the final time measurement for this vs. before all the postprocessing changes?

For the following test (on my machine, run in series under similar conditions):

```python
import timeit

timeit.timeit("data = statcast('2020-08-01', '2020-10-05')", "from pybaseball import statcast", number=1)
```

2.0.0: 99.82 seconds
2.1.0: 253.30 seconds
This: 108.85 seconds

schorrm:
That seems like a wash given variance.

PR #158, opened 2020-10-19
statcast large-query timing script in CI

bdilday:
This PR adds a step to the CI that will execute a large query against statcast and check that it completes within a given time threshold. This will help make sure subsequent PRs are taking care of https://github.com/jldbc/pybaseball/issues/154

### DIFF
--- a/scripts/statcast_timing.py
+++ b/scripts/statcast_timing.py
@@ -16,15 +16,12 @@ def main():
     _ = statcast(START_DATE, END_DATE)
     end_time = time.time()
     query_time = end_time-start_time
+    diagnostic_messages = f"query took {query_time: .1f} seconds which is", f"threshold of {time_threshold: .1f}\n"
     if query_time > time_threshold:
-        sys.stderr.write("query took {:.2f} seconds "
-                         "which is greater than threshold of {:.2f}".format(query_time, time_threshold)
-                         )
+        sys.stderr.write("{0} greater than {1}".format(*diagnostic_messages))
         sys.exit(1)
     else:
-        sys.stdout.write("query took {:.2f} seconds "
-                         "which is less than threshold of {:.2f}".format(query_time, time_threshold)
-                         )
+        sys.stdout.write("{0} less than {1}".format(*diagnostic_messages))
 
 if __name__=='__main__':
     main()
--- a/scripts/statcast_timing.py
+++ b/scripts/statcast_timing.py
@@ -23,7 +23,7 @@ def main():
     end_time = time.time()
     query_time = end_time-start_time
     threshold_exceeded = query_time > args.time_threshold
-    print(f"query took {query_time: .1f} (expected less than {args.time_threshold: .1f})")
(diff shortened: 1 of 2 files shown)


scripts/statcast_timing.py
```diff
@@ -0,0 +1,30 @@
...
+    if len(sys.argv) == 2:
+        time_threshold = float(sys.argv[1])
+    else:
+        time_threshold = EXPECTED_TIME_THRESHOLD
+
+    start_time = time.time()
+    _ = statcast(START_DATE, END_DATE)
+    end_time = time.time()
+    query_time = end_time-start_time
+    if query_time > time_threshold:
```
schorrm, line 19: Do you know if subprocess works in CI? We could run this in a subprocess with a timeout and kill if it went too far over (let's say over a minute).

schorrm:
LGTM though, just waiting to hear your thoughts on capping the runtime for this

scripts/statcast_timing.py
```diff
@@ -0,0 +1,30 @@
...
+        sys.stderr.write("query took {:.2f} seconds "
```
TheCleric, line 20: Should we use f-strings instead of a format call?

scripts/statcast_timing.py
bdilday: no I don't know. the goal is to make CI run faster? it feels like solving a problem we don't have.

scripts/statcast_timing.py
bdilday: yes good idea - I wasn't aware (until I just looked it up now) that you can use a format-code with f-strings.

bdilday: I used part f-strings, part `format`, let me know if loosk o0k

TheCleric: It looks okay. You could also do this:

```python
sys.stdout.write(f"query took {query_time: .1f} seconds which is {"less" if query_time < time_threshold else "greater"} than threshold of {time_threshold: .1f}\n")
```

schorrm: For that '{"less" if query_time < time_threshold else "greater"}'
Should be move outside. No long stuff in there

scripts/statcast_timing.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def main():
+    args = _parse_args()
+    start_time = time.time()
+    _ = statcast(args.start_date, args.end_date)
+    end_time = time.time()
+    query_time = end_time-start_time
+    threshold_exceeded = query_time > args.time_threshold
+    print(f"query took {query_time: .1f} (expected less than {args.time_threshold: .1f})")
```
TheCleric, line 26: Even better! 😄

scripts/statcast_timing.py
bdilday: @schorrm I looked up the [documentation on this](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) and implemented a 1 minute timeout on the large-query step

Issue #156, opened 2020-10-15
Issues with standardizations of names between functions


Am3ra:
The problem here is simple: There are teams like the LA Angels whose acronym in bbref is `LAA`, while in retrosheet and its functions it's `ANA`.

I created the following structures to deal with this problem:

``` python
bbref_current_teams = ["ARI","ATL","BAL","BOS","CHC","CHW","CIN","CLE","COL",
                      "DET","HOU","KCR","LAA","LAD","MIA","MIL","MIN","NYM",
                      "NYY","OAK","PHI","PIT","SDP","SEA","SFG","STL","TBR",
                       "TEX","TOR","WSN"]

retrosheet_current_teams = ["ANA","BAL","BOS","CHA","CLE","DET","HOU","KCA",
                            "MIN","NYA","OAK","SEA","TBA","TEX","TOR","ARI",
                            "ATL","CHN","CIN","COL","LAN","MIA","MIL","NYN",
                            "PHI","PIT","SDN","SFN","SLN","WAS"]

bbref_to_retrosheet = {"ARI":"ARI","ATL":"ATL","BAL":"BAL","BOS":"BOS","CHC":"CHN","CHW":"CHA","CIN":"CIN","CLE":"CLE","COL":"COL",
                      "DET":"DET","HOU":"HOU","KCR":"KCA","LAA":"ANA","LAD":"LAN","MIA":"MIA","MIL":"MIL","MIN":"MIN","NYM":"NYN",
                      "NYY":"NYA","OAK":"OAK","PHI":"PHI","PIT":"PIT","SDP":"SDN","SEA":"SEA","SFG":"SFN","STL":"SLN","TBR":"TBA",
                       "TEX":"TEX","TOR":"TOR","WSN":"WAS"}

retrosheet_to_bbref = {v: k for k, v in bbref_to_retrosheet.items()} 
```

I was wondering if people would be interested in including this in the package, or if something similar already exists? Additionally, this helps to iterate over teams. 

Would like to know your opinions.

TheCleric:
This already exists in the lahman teams data to my knowledge. There should be a fields for `teamIDBR` and `teamIDretro `

schorrm:
This is the sort of thing that could make a **lot** of sense in a `pybaseball.constants`

TheCleric:
@schorrm or we could modify `teamid_lookup.py` to not be specific to Fangraphs and include these other ids as well. Just depends if we want to make it easier, or let Lahman worry about that data.

[...]

PR #159, opened 2020-10-20
Add BRef and Retrosheet to teamid_lookup

TheCleric:
In response to the request that @Am3ra made in #156, this PR adds the BRef and Retrosheet IDs to the teamid_lookup module. As well it changes the name of `fangraphs_teams` to `team_ids` (while maintaining `fangraphs_teams` as an alias for backwards compatibility.


schorrm:
This looks good to me, but honestly, tiny bit worried there's no comments / controversy etc

TheCleric:
> This looks good to me, but honestly, tiny bit worried there's no comments / controversy etc

What do you mean @schorrm ? My PRs NEVER have any comments. They just pop into existence as PERFECTION. Just look at the cache PR. 😉

schorrm:
Lol. In general, if there's nothing, means everybody wasn't paying attn or missed something. I'll try one thing here tho: does this need a docs update?

TheCleric:
> Lol. In general, if there's nothing, means everybody wasn't paying attn or missed something. I'll try one thing here tho: does this need a docs update?

That's actually a good point. I'll take care of that.

PR #162, opened 2020-10-20
Add spell-check suggestions to playerid lookup

tjburch:
Recreation of #161, but instead using chadwick file.

Add spell checking on playerid_lookup. If the passed name to playerid_lookup does not have any matches, the Levenshtein distance is calculated for both the first and last names to a list of all player names, then those distances are summed. If the sum has a single minima, it raises a ValueError and suggests the single minima. If it has two, it will raise the ValueError and suggest both. If more than two, it just raises a ValueError with no suggestion.

Note, previously playerid_lookup did not raise any errors if values were not found, it would only return an empty DF, so the raised errors will be new.

### Example output

One minima:

```
>>> playerid_lookup("turner", "justn")
Gathering player lookup table. This may take a moment.
  File "/Users/tburch/Documents/github/pybaseball/pybaseball/playerid_lookup.py", line 110, in playerid_lookup
    raise ValueError(f"Player not found! Did you mean {suggested_first} {suggested_last}?")
ValueError: Player not found! Did you mean Justin Turner?
```

Two minima:

```
>>> playerid_lookup("Carpenter","")
Gathering player lookup table. This may take a moment.
  File "/Users/tburch/Documents/github/pybaseball/pybaseball/playerid_lookup.py", line 115, in playerid_lookup
    raise ValueError(f"Player not found! Perhaps you meant {suggested_first[0]} {suggested_last[0]} or {suggested_first[1]} {suggested_last[1]}?")
ValueError: Player not found! Perhaps you meant Bob Carpenter or Lew Carpenter?
```

This works really well for simple mistyped names - a few letters messed up. For strange cases, like the second example here, it will preferentially go to shorter names since it's optimizing a distance - at some point it might be worth considering a better ad-hoc rule (e.g. most recent player with distance less than some value), but I think this really works for all reasonable intents.


pybaseball/analysis/string_similarity.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+
+def levenshtein(seq1: str, seq2: str) -> np.int64:
```
bdilday, line 3: I think it'd be better to use a library such as [fuzzywuzzy](https://github.com/seatgeek/fuzzywuzzy) instead of writing a custom levenshtein function. is there a reason to use the custom version here?

pybaseball/playerid_lookup.py
```diff
@@ -52,7 +52,22 @@ def get_lookup_table(save=False):
...
+    Args:
+        last (str): Provided last name
+        first (str): Provided first name
+        player_table (pd.DataFrame): Chadwich player table including names
+
+    Returns:
+        pd.DataFrame: player table with levenshtein distance added
+    """
+    player_table["lastname_dist"] = player_table.dropna().apply(lambda row : levenshtein(row["name_last"], last),axis=1)
+    player_table["firstname_dist"] = player_table.dropna().apply(lambda row : levenshtein(row["name_first"],first),axis=1)
```
bdilday, line 67: `fillna` might be better than `dropna` to allow last name only searches?

pybaseball/playerid_lookup.py
```diff
@@ -52,7 +52,22 @@ def get_lookup_table(save=False):
...
 
+    Args:
+        last (str): Provided last name
+        first (str): Provided first name
+        player_table (pd.DataFrame): Chadwich player table including names
+
+    Returns:
+        pd.DataFrame: player table with levenshtein distance added
+    """
+    player_table["lastname_dist"] = player_table.dropna().apply(lambda row : levenshtein(row["name_last"], last),axis=1)
```
bdilday, line 66: this adds new rows to the `player_table` but it'd be better if this function doesn;t mutate the data

pybaseball/analysis/string_similarity.py
tjburch: I guess that's up to preference, I figured adding another dependency for a ~20 line function would be overkill but if we'd rather use an external library, that's fine with me.

pybaseball/playerid_lookup.py
```diff
@@ -52,7 +52,22 @@ def get_lookup_table(save=False):
...
+    player_table["firstname_dist"] = player_table.dropna().apply(lambda row : levenshtein(row["name_first"],first),axis=1)
+    player_table["total_dist"] = player_table["lastname_dist"] + player_table["firstname_dist"]
```
bdilday, line 68: I don't know if adding them is the best here. maybe use a harmonic mean instead or something?

pybaseball/playerid_lookup.py
```diff
@@ -84,6 +99,24 @@ def playerid_lookup(last=None, first=None, player_list=None):
 
     results = results.reset_index(drop=True)
 
+    if len(results) == 0:
```
bdilday, line 102: this seems complicated enough that it could go in its own function, `get_similar_matches` or something.

pybaseball/playerid_lookup.py
```diff
@@ -84,6 +99,24 @@ def playerid_lookup(last=None, first=None, player_list=None):
...
+        similarity_df = name_similarity(last=last, first=first, player_table=table)
+        minimum_distance = similarity_df[similarity_df["total_dist"] == similarity_df["total_dist"].min()]
+
+        # If there's only one similar name, return that
+        if len(minimum_distance) == 1:
+            suggested_first = minimum_distance["name_first"].iloc[0].capitalize()
+            suggested_last = minimum_distance["name_last"].iloc[0].capitalize()
+            raise ValueError(f"Player not found! Did you mean {suggested_first} {suggested_last}?")
```
bdilday, line 110: I'm not sure if raising a value error is the best thing here. I'd prefer returning an empty result. What's the rationale to raise an error?

pybaseball/analysis/string_similarity.py
schorrm: This looks like a potential perf-killer, do you have measurements on this?

pybaseball/playerid_lookup.py
tjburch: I prefer the notification when the query fails vs silent failure. If the preference is for no errors, fine, but then the question is what the preferred method of return is? I could see:

1) Print guesses to screen and also return the empty DataFrame
2) Show the nearest X players to the DataFrame with a print statement of "no identical matches found, closest X returned"

bdilday: I think we;re mixing up two concepts here, lookup vs search. thinking it through now, I'd expect a lookup with a missing key to raise an error and a search that has no results to return an empty result. so I don;t know where that leaves us - any other thoughts?

tjburch: Probably we should think of the use-case. Personally, when I use this function, I'm starting up an analysis and just trying to find keys for a player I want to study. This means I'm typically working in a Jupyter notebook. In that case, errors are ugly and generally not useful, imo. However, an empty data frame _also_ isn't useful (also it's super frustrating because this function takes some time to get the data, and I know I have to run it again).

If that's the main and/or only use-case, I think solution 2 above works best, showing the X closest players to the entered name with some warning print statement. I just finished writing this code (w/fuzzywuzzy) so that's workable.

However, if you can think of use-cases in which having the empty dataframe or an error message is better for the user, we should probably default to that. None immediately come to mind, but ofc my scope is limited to my experience.

tjburch: >However, if you can think of use-cases in which having the empty dataframe or an error message is better for the user, we should probably default to that. None immediately come to mind, but ofc my scope is limited to my experience.

I just thought some more about this - I could imagine someone doing something conditional on the size of the returned df. Maybe the best option is to add an optional boolean argument `search`, when True it will return the 5 nearest suggestions? We could default it to False so it doesn't disrupt current uses. The only con I see with doing that approach is that I don't imagine many people will use it in that case, but oh well.

bdilday:
I know we said we don't want to require tests from PRs but this could really benefit from some tests. In addition to testing desired behaviour, I think it'd actually help decide what we want this function to do / return.

pybaseball/playerid_lookup.py
tjburch: fixed in commit e0fe3d0

pybaseball/playerid_lookup.py
tjburch: fixed in commit e0fe3d0

pybaseball/playerid_lookup.py
tjburch: switched to fuzzywuzzy process; no longer need to perform this operation

pybaseball/playerid_lookup.py
tjburch: simplified, now just a couple of lines

tjburch:
The new behavior after this commit doesn't disrupt the previous behavior but allows for a search flag to be passed:

```
>>> playerid_lookup("carpenter","mat")
Gathering player lookup table. This may take a moment.
Empty DataFrame
Columns: [name_last, name_first, key_mlbam, key_retro, key_bbref, key_fangraphs, mlb_played_first, mlb_played_last]
Index: []
>>> playerid_lookup("carpenter","mat", search=True)
Gathering player lookup table. This may take a moment.
No identically matched names found! Returning the 5 most similar names.
  name_first  name_last  key_mlbam key_retro  key_bbref  key_fangraphs  mlb_played_first  mlb_played_last
0       matt  carpenter     572761  carpm002  carpema01           8090            2011.0           2020.0
1       paul  carpenter     112023  carpp101  carpepa01        1001981            1916.0           1916.0
2       ryan  carpenter     543001  carpr001  carpery01          12699            2018.0           2019.0
3      bubba  carpenter     150084  carpb001  carpebu01        1001977            2000.0           2000.0
4      david  carpenter     502304  carpd001  carpeda01           3959            2011.0           2019.0
5      david  carpenter     518526  carpd002  carpeda02           9244            2012.0           2015.0
```

bdilday:
> The new behavior after this commit doesn't disrupt the previous behavior but allows for a search flag to be passed:
> 
> ```
> >>> playerid_lookup("carpenter","mat")
> Gathering player lookup table. This may take a moment.
> Empty DataFrame
> Columns: [name_last, name_first, key_mlbam, key_retro, key_bbref, key_fangraphs, mlb_played_first, mlb_played_last]
> Index: []
> >>> playerid_lookup("carpenter","mat", search=True)
> Gathering player lookup table. This may take a moment.
> No identically matched names found! Returning the 5 most similar names.
>   name_first  name_last  key_mlbam key_retro  key_bbref  key_fangraphs  mlb_played_first  mlb_played_last
> 0       matt  carpenter     572761  carpm002  carpema01           8090            2011.0           2020.0
> 1       paul  carpenter     112023  carpp101  carpepa01        1001981            1916.0           1916.0
> 2       ryan  carpenter     543001  carpr001  carpery01          12699            2018.0           2019.0
> 3      bubba  carpenter     150084  carpb001  carpebu01        1001977            2000.0           2000.0
> 4      david  carpenter     502304  carpd001  carpeda01           3959            2011.0           2019.0
> 5      david  carpenter     518526  carpd002  carpeda02           9244            2012.0           2015.0
> ```

🎉 this is pretty nifty! any interest in adding some tests?

pybaseball/playerid_lookup.py
```diff
@@ -52,8 +52,25 @@ def get_lookup_table(save=False):
...
+        pd.DataFrame: 5 nearest matches from fuzzywuzzy.process
+    """
+    filled_df = player_table.fillna("")
+    chadwick_names = filled_df["name_first"] + " " + filled_df["name_last"]
+    fuzzy_matches = pd.DataFrame(process.extract(f"{first} {last}", chadwick_names, limit=5))
+    matched_names = fuzzy_matches[0].str.split(expand=True)
+    matched_names = matched_names.rename(columns={0:"name_first", 1:"name_last"})
+    return matched_names
+    
+def playerid_lookup(last=None, first=None, player_list=None, search=False):
```
bdilday, line 73: I think this needs a doc string, otherwise the cool new feature `search` is obscured

tjburch: Good call, done.

tjburch:
> 🎉 this is pretty nifty! any interest in adding some tests?

@bdilday I'm happy to, though I honestly have no idea where I'd start writing tests for such a function... just baseline checks that with search disabled the length is 0 for invalid inputs, and >0 for valid? Or what sort of tests do you envision here?

bdilday:
> > tada this is pretty nifty! any interest in adding some tests?
> 
> @bdilday I'm happy to, though I honestly have no idea where I'd start writing tests for such a function... just baseline checks that with search disabled the length is 0 for invalid inputs, and >0 for valid? Or what sort of tests do you envision here?

I was thinking tests of the search results, example, 
`playerid_lookup("aron", "hak", search=True`, the first result is Hank Aaron,
`playerid_lookup("hernandez", "kiki", search=True)`, the first result is Kiké Hernández,
`playerid_lookup("martinez", "pedro", search=True)`, gives me two exact matches,
`playerid_lookup("gariciappara", search=True)` gives Nomar Garciaparra,

... any other test cases that seem useful

TheCleric:
> I was thinking tests of the search results, example,
> `playerid_lookup("aron", "hak", search=True`, the first result is Hank Aaron,
> `playerid_lookup("hernandez", "kiki", search=True)`, the first result is Kiké Hernández,
> `playerid_lookup("martinez", "pedro", search=True)`, gives me two exact matches,
> `playerid_lookup("gariciappara", search=True)` gives Nomar Garciaparra,
> 
> ... any other test cases that seem useful

Yep those sound right to me as well. Make sure these go into the integration folder as they are hitting real data. I can assist in making these true unit tests as well.

bdilday:
> `playerid_lookup("hernandez", "kiki", search=True)`, the first result is Kiké Hernández,

this one shouldn't be taken literally, it doesn't have to be this player, I just meant one where the spelling might be tough to guess phonetically and if you take a reasonable guess you should get the player you were looking for back (but this example doesn't illustrate the point because the first name is Enrique in the data)

tjburch:
> this one shouldn't be taken literally, it doesn't have to be this player, I just meant one where the spelling might be tough to guess phonetically and if you take a reasonable guess you should get the player you were looking for back (but this example doesn't illustrate the point because the first name is Enrique in the data)

Yeah, maybe in the long term future we could do some handling for nicknames (I know there's availability on BBref), but that seems like quite a lot for first implementation. I'll try to get some tests up tonight after work.

schorrm:
We should not be doing a pull for each call invocation, if we're already refactoring. Really, you should download the table and then look it up in there.
If that API is an issue, then we should perhaps wrap this in a class and then download only on class instantiation, then we have instantly for subsequent use.

tjburch:
Alright, so I added some tests in the integration folder. @TheCleric, I modeled them after existing tests in that folder, but if you have better suggestions for making them unit tests, let me know.

tjburch:
> We should not be doing a pull for each call invocation, if we're already refactoring. Really, you should download the table and then look it up in there.
> If that API is an issue, then we should perhaps wrap this in a class and then download only on class instantiation, then we have instantly for subsequent use.

This is a good point.... maybe outside the scope of this PR specifically but should be wrapped into the refactor. I can look at that next.

tjburch:
@TheCleric - any suggestions on the tests? Else, I think all the review items are covered here.

TheCleric:
> @TheCleric - any suggestions on the tests? Else, I think all the review items are covered here.

They look pretty good. Only suggestion I would make is splitting each case into its own test, that way we'll know exactly which test case is failing, and if multiple would fail (as opposed to it bailing on the first failed assertion).

tjburch:
Great, thanks! Done in the most recent commit. Otherwise, is there anything else to be done here?

tests/integration/pybaseball/test_playerid_lookup.py
```diff
@@ -0,0 +1,33 @@
...
+    assert gariciappara_df["name_first"][0] == "nomar"
+    assert gariciappara_df["name_last"][0] == "garciaparra"
+
+def test_playerid_lookup_phonetic_yastrzemski() -> None:
+    """Test phonetically tougher names with reasonable guesses (use-case)"""
+    # Mike Yastrezremski
+    yastrzemski_df = playerid_lookup("yastremsky", "mike", search=True)
+    assert yastrzemski_df["name_last"][0] == "yastrzemski"
+
+def test_playerid_lookup_phonetic_yastrzemski() -> None:
```
bdilday, line 28: function name looks like copy and past from above - should be `test_playerid_lookup_phonetic_bogaerts` or something

TheCleric (approved):

schorrm:
I have reservations on hitting the runtime cost of pulling it each time. I would really prefer to see this bundled into some client where instantiation triggers the download process.

schorrm:
I'm not perf obsessed, but this is the difference between several seconds for dling and microsecs or millisecs locally

schorrm:
@tjburch , @bdilday, @TheCleric -- I moved this to a client-based model here, please lmk what you guys think

schorrm:
This is a breaking change, and intentionally so (see the function stubs to return exceptions detailing the deprecation)

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,119 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+        """
+        key_types = (
+            'mlbam',
+            'retro',
+            'bbref',
+            'fangraphs',
+        )
 
+        if not key_type:
+            key_type = key_types[0]  # default is "mlbam" if key_type not provided
```
tjburch, line 151: Would it be more clear to make "mlbam" the default argument in the function declaration?

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,119 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+            results = similar_names_df.merge(self.table, left_on=["name_last", "name_first"], right_on=[
+                                             "name_last", "name_first"], how="left")
+
         return results
 
-    if first is None:
-        results = table.loc[table['name_last'] == last]
-    else:
-        results = table.loc[(table['name_last'] == last) & (table['name_first'] == first)]
+    def reverse_lookup(self, player_ids: List[str], key_type=None) -> pd.DataFrame:
```
tjburch, line 133: Probably we should add a test or two for this when this goes in

tjburch (commented):
Pending comments this looks good, thanks for taking care of migrating it over to a client-approach for me

pybaseball/playerid_lookup.py
schorrm: It would. I migrated existing code pretty thoughtlessly

schorrm: Done, and threw in some F-string updates to there as well

bdilday:
> @tjburch , @bdilday, @TheCleric -- I moved this to a client-based model here, please lmk what you guys think

as a user interface this feels awkward to me. as a user I'd like to do

```
from statcast import player_id_search
player_id = player_id_search(last="yastremski", first="mike")
``` 

and not have to instantiate an object first. the overall goal is to avoid loading the table more than once right? can we accomplish in a different way that doesn't ask the user to import and instantiate an object?

schorrm:
Yeah, we could hide this from the user.
We make a module-wide variable to hold a class object, initialized to None.
We init if None, and then we rely on the module-wide variable

schorrm:
So the actual code being run is what I have here, but the user doesn't see it.

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,116 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+    matched_names = matched_names.rename(columns={0: "name_first", 1: "name_last"})
+    return matched_names
+
+
+class player_search_client:
+    def __init__(self) -> None:
+        self.table = get_lookup_table()
+
+    def search(self, last: str = None, first: str = None, player_list: List[str] = None, fuzzy: bool = False) -> pd.DataFrame:
+        """Lookup playerIDs (MLB AM, bbref, retrosheet, FG) for a given player
```
bdilday, line 83: maybe we could use a pattern like the fangraphs queries, where we instantiate the client object and alias `player_search` to `player_search_client().search`?

https://github.com/jldbc/pybaseball/blob/master/pybaseball/datasources/fangraphs.py#L216

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,116 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+    chadwick_names = filled_df["name_first"] + " " + filled_df["name_last"]
+    fuzzy_matches = pd.DataFrame(process.extract(f"{first} {last}", chadwick_names, limit=5))
+    matched_names = fuzzy_matches[0].str.split(expand=True)
+    matched_names = matched_names.rename(columns={0: "name_first", 1: "name_last"})
+    return matched_names
+
+
+class player_search_client:
+    def __init__(self) -> None:
+        self.table = get_lookup_table()
```
bdilday, line 80: I think it'd be better to initialize this to `None`, then load it lazily the first time the `search` method gets called. 
example, 

```
class player_search_client:
  def __init__(self):
    self._table = None
  @property
  def table(self):
    if self._table is None:
      self._table = get_lookup_table()
   return self._table
```

then in `search` you can reference `self.table` as before

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,116 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+                if (len(player.split(",")) > 1):
+                    first = player.split(",")[1].strip()
+                if (player_counter == 1):
+                    results = self.search(last, first)
+                else:
+                    results = results.append(self.search(last, first), ignore_index=True)
+                player_counter += 1
+            return results
+
+        if first is None:
```
bdilday, line 117: this if else block implies that supplying first name without last name is an invalid usage. should we make `last` a required argument?

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,116 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+            first (str, optional): Player's first name. Defaults to None.
+            player_list (list, optional): List of players to seach for (comma delimited). Defaults to None.
+            search (bool, optional): In case of typos, returns players with names close to input. Defaults to False.
+
+        Returns:
+            pd.DataFrame: DataFrame of playerIDs, name, years played
+        """
+
+        # force input strings to lowercase
+        if last:
```
bdilday, line 96: should `last` be a required argument?

pybaseball/playerid_lookup.py
schorrm: IIRC the FG base class is low-cost and is executed every time. This one would be a module singleton. I think the model here is what we do with the Lahman ZipFile in memory.

schorrm: Simple outline:
```python3
_client = None
def player_search(...):
    global _client
    if not _client:
        _client = player_search_client()
    return _client.search(...)
```

pybaseball/playerid_lookup.py
schorrm: I don't think so since there's the list option

pybaseball/playerid_lookup.py
schorrm: See above, I don't think so since list is possible

pybaseball/playerid_lookup.py
```diff
@@ -47,77 +50,116 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
...
+            last = last.lower()
+        if first:
+            first = first.lower()
+
+        # if player_list has a value, then the user is passing in a list of players
+        # the list of players may be comma delimited for last, first, or just last
+        if player_list:
```
bdilday, line 103: do we need to support a list? I'd suggest just getting rid of this and restore it if /when someone asks for it. alternatively make processing a list an explicitly different method? I basically think the difference in the way the arguments are inputted is confusing, like if you want one player you enter `last=last, first=first`, but if you want a list you enter comma separated strings so I'd prefer they're consistent or a list is a different method

schorrm: True. We could unwind this to a separate method -- perhaps just tuples of the pairs? And we can definitely unwind this.

pybaseball/playerid_lookup.py
bdilday: if I understand what you mean then I think the answer is no, the base class is instantiated only once, because of the way `import` keeps track of what's been imported. example, if you say 
```
from .datasources.fangraphs import fg_batting_data
```

then when you call `fg_batting_data()` it reuses the object it instantiated the first time

bdilday: I think what you suggested is fine, however, the only argument against it would be trying to avoid a `global`, which is mainly an argument about style

schorrm: For simplicity's sake, easier to do this with the global (especially since scope is limited)

schorrm:
NOTE: this still needs a documentation update

tjburch:
Sorry, been a hectic few weeks. Updated the docs, should be straightforward enough, lmk if they're reasonable.

schorrm:
LGTM

PR #164, opened 2020-10-21
Code diet for Statcast + Date Iter

schorrm:
This PR is trying to clean up the Statcast code. Functionality is mostly unchanged, with two exceptions:
1. We now delegate the date thing to a specialized iterator (see `date_range` in utils). I think this cleans stuff up a lot.
2. Instead of a complicated system of backoffs and recoveries, we now approach fault tolerance in large requests by yelling at people to turn their cache on if it is off. And it's probably a more fault tolerant approach anyway, to boot.


pybaseball/utils.py
```diff
@@ -48,6 +48,32 @@ def validate_datestring(date_text: Optional[str]) -> date:
...
+
+    low = start
+
+    while low <= stop:
+        if (low.month, low.day) <= (3, 15):
+            low = low.replace(month=3, day=15)
+            if verbose:
+                print('Skipping offseason dates')
+        elif (low.month, low.day) >= (11, 15):
+            low = low.replace(month=3, day=15, year=low.year + 1)
```
tjburch, line 66: Maybe it makes sense to add some sort of catch for 2020 here? Else it will iterate from March to July with a bunch of empty queries.

pybaseball/statcast.py
```diff
@@ -1,175 +1,70 @@
...
 
     return data
 
 
-def large_request(start_date: date, end_date: date, step: int, verbose: bool,
+_OVERSIZE_WARNING = '''
+That's a nice request you got there. It'd be a shame if something were to happen to it.
+We strongly recommend that you enable caching before running this. It's as simple as `pybaseball.cache.enable()`.
+Since the Statcast requests can take a *really* long time to run, if something were to happen, like: a disconnect;
+gremlins; computer repair by associates of Rudy Giuliani; electromagnetic interference from metal trash cans; etc.;
```
tjburch, line 38: The Rudy Giuliani comment got a laugh out of me, but after news stories today (Oct 21) that might be a name we want to stay away from...

pybaseball/statcast.py
```diff
@@ -1,175 +1,70 @@
...
+you could lose a lot of progress. Enabling caching will allow you to immediately recover all the successful
+subqueries if that happens.'''
+
+def check_warning(start_dt: date, end_dt: date) -> None:
+    if not cache.config.enabled and (end_dt - start_dt).days >= 42:
```
tjburch, line 43: It might be better to think about this in terms of memory rather than days. I could see a case where you trigger this by grabbing the last week of one season and the first week of the next and it still being thrown despite the data covering 2 weeks.

pybaseball/utils.py
schorrm: True

pybaseball/statcast.py
schorrm: Really it should be by ticks of the date range I guess.

schorrm: So I guess I'll switch to this to len() on the range iterator or something. Good call.

pybaseball/utils.py
bdilday: or use the actual start and end date for each season instead of always using march 15 and november 15

pybaseball/statcast.py
```diff
@@ -1,175 +1,70 @@
...
 import pybaseball.datasources.statcast as statcast_ds
 
-from .utils import sanitize_date_range
+from .utils import sanitize_date_range, date_range
+from . import cache
 
 _SC_SINGLE_GAME_REQUEST = "/statcast_search/csv?all=true&type=details&game_pk={game_pk}"
 # pylint: disable=line-too-long
 _SC_SMALL_REQUEST = "/statcast_search/csv?all=true&hfPT=&hfAB=&hfBBT=&hfPR=&hfZ=&stadium=&hfBBL=&hfNewZones=&hfGT=R%7CPO%7CS%7C=&hfSea=&hfSit=&player_type=pitcher&hfOuts=&opponent=&pitcher_throws=&batter_stands=&hfSA=&game_date_gt={start_dt}&game_date_lt={end_dt}&team={team}&position=&hfRO=&home_road=&hfFlag=&metric_1=&hfInn=&min_pitches=0&min_results=0&group_by=name&sort_col=pitches&player_event_sort=h_launch_speed&sort_order=desc&min_abs=0&type=details&"
 _MAX_SC_RESULTS = 40000
```
bdilday, line 15: doesn't seem like `_MAX_SC_RESULTS` is being used anymore? we should probably still catch the case where the query returns exactly 40000 rows because it's incomplete and basically a failure to complete the request

pybaseball/statcast.py
```diff
@@ -1,175 +1,70 @@
...
+gremlins; computer repair by associates of Rudy Giuliani; electromagnetic interference from metal trash cans; etc.;
+you could lose a lot of progress. Enabling caching will allow you to immediately recover all the successful
+subqueries if that happens.'''
+
+def check_warning(start_dt: date, end_dt: date) -> None:
+    if not cache.config.enabled and (end_dt - start_dt).days >= 42:
+        warnings.warn(_OVERSIZE_WARNING)
+
+
+def large_request(start_dt: date, end_dt: date, step: int, verbose: bool,
```
bdilday, line 47: we migt want to name this `_large_request` since it's not really meant to be used directly

pybaseball/utils.py
schorrm: Do you know where I can get that? At least for since 2008, which is when the Statcast search starts (based on PitchFX I believe?)

pybaseball/statcast.py
schorrm: Would 'Rudolph W. Giuliani' be better?

pybaseball/statcast.py
schorrm: Throw a highest level warning hoping that someone will report it as a GH issue and then we crank down our date chunking?
Parenthetical: the backoff was problematic, and the efficiency hit of asking, getting 40k, and then redoing it is severe. I am okay with not being fault tolerant in that sense where arguably it _should_ break. I don't see anything wrong with breaking on a breaking API change upstream. It's natural.

pybaseball/statcast.py
schorrm: Or even `_delegate_request` or something since this is now an unconditional code path

pybaseball/utils.py
schorrm: If we can put this in as a hardcoded dict, that'd be ideal.

pybaseball/statcast.py
schorrm: I'm having second thoughts about this, given that it's a usage warning etc.

pybaseball/statcast.py
bdilday: I think we can assume we won't exceed the 40k and just remove the unused constant `_MAX_SC_RESULTS`

pybaseball/utils.py
TheCleric: I think we'd just have to find opening day/final day for each season manually and add a dict.

pybaseball/statcast.py
TheCleric: "Former Mayors of New York City"

pybaseball/statcast.py
TheCleric: I agree with @bdilday especially since we backed it to 7 days per chunk max.

pybaseball/statcast.py
```diff
@@ -1,175 +1,70 @@
...
-            error_counter = 0
-            break
-
-    if dataframe_list:
-        return pd.concat(dataframe_list, axis=0)
-
-    return None
-
-
+@cache.df_cache(expires=365)
```
TheCleric, line 18: I don't think you need this here. The function `get_statcast_data_from_csv_url` from `datasources/statcast.py` has a cache on it already, and this calls that function. I would just add the expiration to that existing wrapper.

pybaseball/utils.py
TheCleric: Looks like BRef has a page we could scrape once and add to a dict (or CSV):

https://www.baseball-reference.com/leagues/MLB/2019-schedule.shtml

schorrm: Dict. I want this to be cheap enough to be used as a normal iterator.

pybaseball/statcast.py
schorrm: I wanted this here for clarity really, didn't want to run it thru another indirection layer before hitting the cache.

pybaseball/statcast.py
schorrm: I kind of want to leave it in a comment?

schorrm: It's good to know / have there

TheCleric: That's a good point. Should we just add a small check to this function where if length of results happens to hit this, we just throw an error or warning?

pybaseball/utils.py
bdilday: https://github.com/jldbc/pybaseball/pull/165

pybaseball/statcast.py
schorrm: I like it as-is, praise be Borat

schorrm:
@bdilday this makes a change to your dict, it's now dict of tuples instead of dict of dicts to make the code paths a bit simpler, if that's okay?

pybaseball/utils.py
```diff
@@ -90,6 +91,35 @@ def date_range(start: date, stop: date, step: int, verbose: bool = True) -> Iter
...
+    '''
+    Iterate over dates. Skip the offseason dates. Returns a pair of dates for beginning and end of each segment.
+    Range is inclusive of the stop date.
+    If verbose is enabled, it will print a message if it skips offseason dates.
+    This version is Statcast specific, relying on skipping predefined dates from STATCAST_VALID_DATES.
+    '''
+    low = start
+
+    while low <= stop:
+        year_start, year_end = STATCAST_VALID_DATES.get(low.year,
```
bdilday, line 104: suggest naming this `season_start`, `season_end` instead of `year_start`, `year_end`

schorrm:
Since yesterday was last day, I tweaked to that day. @bdilday , final sign-off?

PR #165, opened 2020-10-22
statcast valid dates

bdilday:
A script to determine valid dates for statcast queries. Also included is the resulting json which can be used in `date_range` to exclude invalid dates on a season by season basis.


schorrm:
I think this makes sense in dict form, with adjustment for 2020

bdilday:
> I think this makes sense in dict form, with adjustment for 2020

sounds good. do we need season start and end dates for pre-2008? I don't immediately see a reason.

pybaseball/utils.py
```diff
@@ -40,6 +40,22 @@
...
+    2011: {'min_valid_date': '2011-03-31', 'max_valid_date': '2011-10-28'},
+    2012: {'min_valid_date': '2012-03-28', 'max_valid_date': '2012-10-28'},
+    2013: {'min_valid_date': '2013-03-31', 'max_valid_date': '2013-10-30'},
+    2014: {'min_valid_date': '2014-03-22', 'max_valid_date': '2014-10-29'},
+    2015: {'min_valid_date': '2015-04-05', 'max_valid_date': '2015-11-01'},
+    2016: {'min_valid_date': '2016-04-03', 'max_valid_date': '2016-11-02'},
+    2017: {'min_valid_date': '2017-04-02', 'max_valid_date': '2017-11-01'},
+    2018: {'min_valid_date': '2018-03-29', 'max_valid_date': '2018-10-28'},
+    2019: {'min_valid_date': '2019-03-20', 'max_valid_date': '2019-10-30'},
+    2020: {'min_valid_date': '2020-07-23', 'max_valid_date': '2020-11-15'}
```
schorrm, line 56: Make max valid for this year like 10-30 or something?

bdilday: I figured we'd update it when the world series is over

schorrm: Eh, let's go for a somewhat tighter upper bound (and now we should know to within a day really).

schorrm:
Can I just change the dtypes here to dates?
I'll merge then, I'll also update #164 to use this instead (perhaps splitting off a statcast_date_range from a general baseball_date_range?)

PR #167, opened 2020-10-30
make the large query print only if more than step

schorrm:
Simple thingy, just check that the two days are less than step days apart.


TheCleric:
The code itself looks fine, but I'm concerned with the fact the tests are failing all of a sudden...

schorrm:
Doesn't seem related at all 🤷‍♂️

schorrm:
Closing this because of the gremlin in the tests and also because #168 overrides this sort of thing

PR #168, opened 2020-10-30
Parallelize statcast calls

TheCleric:
In this PR I've done a few things:
* Create a ProcessPoolExecutor to run statcast calls in parallel
* Change the statcast split to a single day, so that if the cache is enabled, each day is cached separately for ease of data reuse for overlapping calls
* Also I added a pytest plugin to parallelize the tests. They now run much faster.

The speed effects for this are pretty dramatic. For the following code:
```python
import timeit

timeit.timeit("data = statcast('2020-08-01', '2020-10-05')", "from pybaseball import statcast", number=1)
```

We've gone from @ 108 seconds to @ 26 seconds. So roughly **4x** faster.

Also the tests have gone from @ 350 seconds to @ 140 seconds, so @ **2.5x** faster.

So there are some downsides:
* The print statements within the workers seem to disappear. Not sure what can be done about this. Since the results are now much faster, this is slightly less of an issue, but for EXTREMELY large data sets (for example 10 seasons of data), it can still take a while with no user feedback. We could perhaps give an estimate at the beginning of how long we think the call may take to set expectations ahead of time. Or maybe since each season is pretty fast, do a loop of each season serially, and data within it in parallel, giving a print as we finish each season.
* The pytest data injection/monkeypatch does not seem to work with these parallel tasks, so I've had to remove the unit testing around the statcast call. We still have integration tests for these though.

### DIFF
--- a/pybaseball/statcast.py
+++ b/pybaseball/statcast.py
@@ -4,6 +4,7 @@ from datetime import date
 from typing import Optional, Union
 
 import pandas as pd
+from tqdm import tqdm
 
 import pybaseball.datasources.statcast as statcast_ds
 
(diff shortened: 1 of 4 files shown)


schorrm:
How do we feel about p_tqdm?

schorrm:
Moving to one call per day -- what does that do to our performance?

schorrm:
Okay -- we shouldn't have both tqdm **and** print statements with updates.
Even `tqdm.write` will look clunky given the number of statements

pybaseball/statcast.py
```diff
@@ -39,35 +41,41 @@ def _small_request(start_dt: date, end_dt: date, team: Optional[str] = None, ver
...
     Fulfill the request in sensible increments.
     """
 
     _check_warning(start_dt, end_dt)
 
-    dataframe_list = []
-
     if verbose:
-        print("This is a large query, it may take a moment to complete")
+        print("This is a large query, it may take a moment to complete\n", flush=True)
```
schorrm, line 59: do we need the trailing \n here?

PR #169, opened 2020-11-12
Spell Check bug fix for 3 word names

mwisnie5:
This PR resolves a bug in the new spell check changes for playerid_lookup that isn't handling names that are abbreviated or three word names like JD Martinez, Hyun Jin Ryu, or Tommy La Stella.

### DIFF
--- a/pybaseball/playerid_lookup.py
+++ b/pybaseball/playerid_lookup.py
@@ -107,8 +107,9 @@ class _PlayerSearchClient:
         # If no matches, return 5 closest names
         if len(results) == 0 and fuzzy:
             print("No identically matched names found! Returning the 5 most similar names.")
+            results=get_closest_names_df(last=last, first=first, player_table=self.table)
             
-        return get_closest_names_df(last=last, first=first, player_table=self.table)
+        return results
 
 
     def search_list(self, player_list: List[Tuple[str, str]]) -> pd.DataFrame:
--- a/pybaseball/playerid_lookup.py
+++ b/pybaseball/playerid_lookup.py
@@ -58,12 +58,14 @@ def get_lookup_table(save=False):
     return table
 
 
-def name_similarity(last: str, first: str, player_table: pd.DataFrame) -> pd.DataFrame:
+def get_closest_names_df(last: str, first: str, player_table: pd.DataFrame) -> pd.DataFrame:
     """Calculates similarity of first and last name provided with all players in player_table
+
     Args:
         last (str): Provided last name
         first (str): Provided first name
         player_table (pd.DataFrame): Chadwick player table including names
+
     Returns:
         pd.DataFrame: 5 nearest matches from fuzzywuzzy.process
     """
@@ -72,7 +74,7 @@ def name_similarity(last: str, first: str, player_table: pd.DataFrame) -> pd.Dat
     fuzzy_matches = pd.DataFrame(process.extract(f"{first} {last}", chadwick_names, limit=5))
     fuzzy_indices = fuzzy_matches[2].tolist()
 
-    return fuzzy_indices
+    return filled_df.iloc[fuzzy_indices].reset_index(drop=True)
 
 
 class _PlayerSearchClient:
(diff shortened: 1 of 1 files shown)


pybaseball/playerid_lookup.py
```diff
@@ -30,16 +30,20 @@ def chadwick_register(save: bool = False) -> pd.DataFrame:
 
     print('Gathering player lookup table. This may take a moment.')
     s = requests.get(url).content
-    mlb_only_cols = ['key_retro', 'key_bbref', 'key_fangraphs', 'mlb_played_first', 'mlb_played_last']
+    mlb_only_cols = ['key_retro', 'key_bbref',
```
bdilday, line 33: there are a lot of formatting changes in this PR that don't relate to the bug. it'd be easier to understand what caused the bug and which changes affect it if there were less formatting changes.

mwisnie5: Oh, yea. Sorry about that. I had an auto-formatter turned on. I'll update shortly with it turned off.

mwisnie5: Updated in latest checkin

pybaseball/playerid_lookup.py
```diff
@@ -108,19 +104,16 @@ def search(self, last: str, first: str = None, fuzzy: bool = False) -> pd.DataFr
         if len(results) == 0 and fuzzy:
             print("No identically matched names found! Returning the 5 most similar names.")
             similar_names_df = name_similarity(last=last, first=first, player_table=self.table)
-            results = similar_names_df.merge(self.table, left_on=["name_last", "name_first"], right_on=[
-                                             "name_last", "name_first"], how="left")
+            results = self.table.iloc[similar_names_df].reset_index(drop=True)
```
bdilday, line 107: we have `name_similarity` returning a list of indexes, and then we subset with `iloc` and return that. It seems like we don;t need two steps here, `name_similarity` could do the subsetting and return the result dataframe?

i.e., 
```
return name_similarity(last=last, first=first, player_table=self.table)
```

pybaseball/playerid_lookup.py
```diff
@@ -60,21 +60,19 @@ def get_lookup_table(save=False):
...
     Returns:
         pd.DataFrame: 5 nearest matches from fuzzywuzzy.process
     """
     filled_df = player_table.fillna("")
     chadwick_names = filled_df["name_first"] + " " + filled_df["name_last"]
     fuzzy_matches = pd.DataFrame(process.extract(f"{first} {last}", chadwick_names, limit=5))
-    matched_names = fuzzy_matches[0].str.split(expand=True)
-    matched_names = matched_names.rename(columns={0: "name_first", 1: "name_last"})
-    return matched_names
+    fuzzy_indices = fuzzy_matches[2].tolist()
```
bdilday, line 73: same comment as below, instead of returning the indexes, how about subsetting the dataframe and returning that?
i.e.
```
return filled_df.iloc[fuzzy_indices].reset_index(drop=True)
```

pybaseball/playerid_lookup.py
```diff
@@ -60,21 +60,19 @@ def get_lookup_table(save=False):
 
 def name_similarity(last: str, first: str, player_table: pd.DataFrame) -> pd.DataFrame:
```
bdilday, line 61: what this function does has changed so the name should really change too

schorrm:
Thanks for the fix, @mwisnie5!
1. Can you address @bdilday's comment?
2. Can you undo the formatting diffs (still a lot, esp in docstrings)

pybaseball/playerid_lookup.py
mwisnie5: Done

pybaseball/playerid_lookup.py
mwisnie5: done

pybaseball/playerid_lookup.py
mwisnie5: I changed the name to get_closest_names_df but open to any feedback or other suggestions.

mwisnie5:
Looking into the failed check now.

mwisnie5:
The "test_call_cache_save_fails_silently " test failed, but after taking a look at this, I'm struggling to understand what this test is doing or how it is related to the changes I made. Can anyone give me some help on this one?

schorrm:
I don't see how this is related either. @TheCleric ???

bdilday:
is it at all possible that the parallel pytest makes this non-determinstic? can you remove the `-n 5` flag from the github workflow and see if it resolves it?

TheCleric:
> is it at all possible that the parallel pytest makes this non-determinstic? can you remove the `-n 5` flag from the github workflow and see if it resolves it?

Always possible but each test was designed to run in isolation. So in theory they can run in any order. We could try and see. However I do remember seeing a random failure on one of @schorrm 's PRs even before I added the parallel testing. Can we simply rerun the tests?

PR #170, opened 2020-11-20
Tests update for playerid lookup changes

mwisnie5:


tests/integration/pybaseball/test_playerid_lookup.py
```diff
@@ -30,4 +30,25 @@ def test_playerid_lookup_phonetic_bogaerts() -> None:
...
-    assert bogaerts_df["name_first"][0] == "xander"
\ No newline at end of file
+    assert bogaerts_df["name_first"][0] == "xander"
+    
+def test_playerid_lookup_three_word_name() -> None:
+    """Test names with three words in them"""
+    # Hyun Jin Ryu
+    ryu_df = playerid_lookup("jin ryu", "hyun", fuzzy=True)
+    assert ryu_df["name_last"][0] == "ryu"
+    assert ryu_df["name_first"][0] == "hyun jin"
```
tjburch, line 40: I notice Hyun-jin Ryu is hyphenated on FanGraphs, but in Chadwick it must be three names. Should we add another test for hyphenations?

```
>>> playerid_lookup("Kiner Falefa", "isiah", fuzzy=True)
Gathering player lookup table. This may take a moment.
No identically matched names found! Returning the 5 most similar names.
      name_last name_first  key_mlbam key_retro  key_bbref  key_fangraphs mlb_played_first mlb_played_last
0  kiner-falefa      isiah     643396  kinei001  kineris01          16512             2018            2020
```

schorrm: keep in mind that `mlb_played_last` will likely change though
Actually, if we could drop `mlb_played_last` from the DF or change _all_ of the players in the tests to retired that'd be nice. Because as it stands this is a ticking time bomb for the testing. @mwisnie5

mwisnie5: I made the changes to add the Isiah Kiner-Falefa test. @schorrm I didn't make any changes to the mlb_played_last, however. That field shouldn't make the test fail and its a useful value to have returned from the chadwick register to find the right guy if there are multiple people with the same name.

schorrm:
Fair enough. It’s good code to have in there, no doubt, just wanted to
check we weren’t testing dataframe equality.

tests/integration/pybaseball/test_playerid_lookup.py
```diff
@@ -30,4 +30,33 @@ def test_playerid_lookup_phonetic_bogaerts() -> None:
...
+    """Test names with abbreviations in them"""
+    # Ronald Acuna Jr
+    acuna_df = playerid_lookup("acuna jr.", "ronald", fuzzy=True)
+    assert acuna_df["name_last"][0] == "acuna"
+    assert acuna_df["name_first"][0] == "ronald"
+    
+def test_playerid_lookup_hyphenated_name() -> None:
+    """Test names with abbreviations in them"""
+    # Isiah Kiner-Falefa
+    acuna_df = playerid_lookup("Kiner Falefa", "isiah", fuzzy=True)
```
bdilday, line 59: `acuna_df` doesn't look like an accurate name for this test, could you change it?

mwisnie5: My bad. Just pushed the changes.

PR #174, opened 2020-12-24
Add Marcels Example

andersonfrailey:
This PR adds a very simple example for using the Marcels projection system in pybaseball. Let me know if there's more y'all would be interested in me adding.


schorrm:
I think this is a good idea

PR #176, opened 2021-01-26
Fix BREF standings code; fix statcast exitvelo test

TheCleric:


pybaseball/utils.py
```diff
@@ -64,8 +64,19 @@ def validate_datestring(date_text: Optional[str]) -> date:
     except (AssertionError, ValueError):
         raise ValueError("Incorrect data format, should be YYYY-MM-DD")
 
+def most_recent_season() -> int:
```
bdilday, line 67: you're calling this function in a lot of different places, how about adding an `functools.lru_cache` decorator?

pybaseball/utils.py
```diff
@@ -64,8 +64,19 @@ def validate_datestring(date_text: Optional[str]) -> date:
...
+    '''
+    Find the most recent season.
+    
+    Will be either this year (if the season has started or just ended)
+    or last year (if the season has not yet started).
+    '''
+
+    dates = list(date_range((datetime.today() - timedelta(weeks=52)).date(), datetime.today().date(), verbose=False))
```
schorrm, line 75: Can you please explain this?

TheCleric: Should be a little more clear now.

PR #177, opened 2021-02-12
replace fuzzywuzzy

derrickknox:
FuzzyWuzzy is GPL Licensed, which is incompatible with the license used by this project. This Pull Request replaces FuzzyWuzzy with [RapidFuzz](https://github.com/maxbachmann/rapidfuzz), which implements the same algorithms. It is licensed in a compatible way (MIT License) and is faster than FuzzyWuzzy.

### DIFF
--- a/pybaseball/playerid_lookup.py
+++ b/pybaseball/playerid_lookup.py
@@ -5,7 +5,7 @@ from typing import List, Tuple
 
 import pandas as pd
 import requests
-from fuzzywuzzy import process
+from rapidfuzz import process
 
 from . import cache
 
@@ -67,7 +67,7 @@ def get_closest_names(last: str, first: str, player_table: pd.DataFrame) -> pd.D
         player_table (pd.DataFrame): Chadwick player table including names
 
     Returns:
-        pd.DataFrame: 5 nearest matches from fuzzywuzzy.process
+        pd.DataFrame: 5 nearest matches from rapidfuzz.process
     """
     filled_df = player_table.fillna("")
     chadwick_names = filled_df["name_first"] + " " + filled_df["name_last"]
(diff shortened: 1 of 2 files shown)


cro1100:
I listed this question under issues and thought I would add it as a comment to this most recent thread.
Hi, I have 2 things:
1.) I'm a newbie to data engineering, but I'd like to get involved as a contributor. As I look at the documentation, I understand how it works, but as I try to dig into the code, I don't think I can see anything. Is there something which needs to be done for me to see more of this? What would you like from me?
2.) I'm working on a class project and noticing when I pull season stats, it only pulls "qualified players". I'd love to be able to grab data based on a lower limit of innings/plate appearances and am wondering if there's a way to pull fangraphs data using that additional filter. Is there a way to do this?

I love what has been done here and hope I can add something to the project! Thanks!

schorrm:
@cro1100 -- I responded to your issue -- please stick to there

schorrm:
@derrickknox We don't use FuzzyWuzzy heavily so a performance benefit doesn't really outweigh the downsides of swapping to a more uncommon dependency at this point.
With respect to the license: the GPL documentation indicates that MIT/Expat is compatible with the GPL -- https://www.gnu.org/licenses/license-list.html#Expat

derrickknox:
> With respect to the license: the GPL documentation indicates that MIT/Expat is compatible with the GPL -- https://www.gnu.org/licenses/license-list.html#Expat

This is a compatibility in one direction. So you can use MIT licensed code in GPL licensed code, but not the other way around.

schorrm:
Upon reviewing yes, you are correct, though I do not find the runtime thing totally convincing. However, as I wasn't thrilled about the dependency in the first place, I'm going to be closing this in favor of #180 .

PR #182, opened 2021-02-16
Fix statcast exitvelo tests, and use our own qualifier

TheCleric:


schorrm:
I'm just happy to see a green check for a change

PR #187, opened 2021-03-06
Statcast pitcher leaderboards

wfordh:
Code, docs, and tests for several of the [Statcast pitcher leaderboards](https://baseballsavant.mlb.com/leaderboard/statcast?type=pitcher).


wfordh:
I think this is blocked on the Fangraphs PR (and maybe not best practice to fix statcast batter tests here), but once that's good then this should be, too.

tests/integration/pybaseball/test_statcast_pitcher.py
```diff
@@ -1,11 +1,91 @@
...
+
+def test_statcast_pitcher_exitvelo_barrels() -> None:
+	minBBE = 100
+	result: pd.DataFrame = statcast_pitcher_exitvelo_barrels(2019, minBBE=minBBE)
+
+	assert result is not None
+	assert not result.empty
+
+	assert len(result.columns) == 19
+	assert len(result) == 432
```
schorrm, line 32: We've had issues with asserting exact lengths before when SC just changes the qualifying parameters before, I think we shouldn't have this in the test suite (that was what broke batting)

wfordh: Yeah that's why I specified the minimum number of events for each test, but happy to remove it or change it to just > 0 or something

wfordh:
I don't love how Statcast isn't uniform with how they describe pitches, ie for the spin comp leaderboard the pitches are spelled out, but in pitch movement they are abbreviated. Do we want to standardize them ourselves? And do we want to rename `statcast_pitcher_spin_dir_comp()` to spell it all out or would that be too long? `statcast_pitcher_spin_direction_comparison()`

tests/integration/pybaseball/test_statcast_pitcher.py
schorrm: Yeah but we had a break with minBBE too

schorrm:
@wfordh abbreviated is okay
and standardizing pitches makes sense

wfordh:
Cool. Any preference on using the abbreviations for the pitches or the entire name as the standards? If not, I'll just go with whatever is easier to implement.

schorrm:
I would suggest accepting both across the board

pybaseball/statcast_pitcher.py
```diff
@@ -87,3 +88,19 @@ def statcast_pitcher_percentile_ranks(year: int) -> pd.DataFrame:
...
     # URL returns a null player with player id 999999, which we want to drop
     return data.loc[data.player_name.notna()].reset_index(drop=True)
+
+@cache.df_cache()
+def statcast_pitcher_spin_dir_comp(year: int, pitch_combo: str = "4 - Seamer / Changeup", minP: int = 100, pov: str = "Pit") -> pd.DataFrame:
+    pitches = ["4-Seamer", "Changeup", "Curveball", "Slider", "Sinker", "Cutter"]
+    pitch_combos = [' / '.join(perm) for perm in permutations(pitches, 2)]
+    if pitch_combo not in pitch_combos:
+        print(f"Not a valid pitch combo. Must include two of {pitches}, separated by ' / '.")
+        return
```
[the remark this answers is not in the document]
TheCleric: As well I'd suggest doing all string comparisons as lower case. No sense in erroring if someone used `curveball` instead of `Curveball`

pybaseball/statcast_pitcher.py
```diff
@@ -87,3 +88,19 @@ def statcast_pitcher_percentile_ranks(year: int) -> pd.DataFrame:
...
+    povs = ["Pit", "Bat"]
+    if pov not in povs:
+        print(f"Not a valid pov. Must be one of {', '.join(povs)}")
+        return
```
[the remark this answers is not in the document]
TheCleric: Or an Enum?

** ducks and runs away before @schorrm throws something at me **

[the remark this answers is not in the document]
schorrm: @TheCleric  I was actually thinking about it

pybaseball/statcast_pitcher.py
```diff
@@ -25,3 +29,78 @@ def statcast_pitcher(start_dt: Optional[str] = None, end_dt: Optional[str] = Non
     df = split_request(start_dt, end_dt, player_id, url)
 
     return df
+
+@cache.df_cache()
+def statcast_pitcher_exitvelo_barrels(year: int, minBBE: Union[int, str] = "q") -> pd.DataFrame:
+    url = f"https://baseballsavant.mlb.com/leaderboard/statcast?type=pitcher&year={year}&position=&team=&min={minBBE}&csv=true"
+    res = requests.get(url, timeout=None).content
```
TheCleric, line 36: This isn't a problem with this PR specifically, but we're doing the same thing over and over here and in statcast_batter. I know I made `get_statcast_data_from_csv_url` in `datasources.statcast` but that uses the `try_parse_dataframe` that seems to slow things down quite a bit. I would suggest that we make a new function there like `get_statcast_data_from_csv_url_raw` that does all of this repetitive work for us without the overhead of `try_parse_dataframe`.

tests/integration/pybaseball/test_statcast_pitcher.py
```diff
@@ -22,14 +22,15 @@ def test_statcast_pitcher() -> None:
     assert len(result) == 1982
 
 def test_statcast_pitcher_exitvelo_barrels() -> None:
-	minBBE = 100
-	result: pd.DataFrame = statcast_pitcher_exitvelo_barrels(2019, minBBE=minBBE)
+	min_bbe = 100
+	result: pd.DataFrame = statcast_pitcher_exitvelo_barrels(2019, minBBE=min_bbe)
```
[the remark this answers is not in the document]
TheCleric: The latter is definitely the more Pythonic way.

pybaseball/statcast_pitcher.py
schorrm: I would suggest a shorter name if you decide to go that route

schorrm: Can we just use `requests.get(url, kwargs)` for the other parameters?

TheCleric: Only if you removed them from the original url and named them exactly as they appear in the query strings I believe.

pybaseball/statcast_pitcher.py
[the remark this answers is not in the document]
wfordh: Cool. The returned data has `pitch_a` and `pitch_b`, so I'll use that nomenclature and will do the string comparisons in lowercase. Good call there

pybaseball/statcast_pitcher.py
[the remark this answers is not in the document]
wfordh: I'm not familiar with Enums, but yeah it could be a boolean and just set the default to whatever they use on the site.

pybaseball/statcast_pitcher.py
wfordh: I did mess around with using a `base_url = f"https://baseballsavant.mlb.com/leaderboard/{leaderboard_name}"` variable and then passing a `params = {"type": "pitcher", "minBBE": "q"...}` dict to the requests call, ie `requests.get(base_url, params=params)`. This would probably be more flexible and would maybe allow us to basically bring all of the statcast functions into one file where the names aren't so long, but IDK if I really want to embark on a refactor that could break people's code. 

I'll check out the stuff in `datasources.statcast`

pybaseball/statcast_pitcher.py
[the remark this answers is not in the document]
schorrm: Actually @TheCleric I misread this from email, I thought this about an enum in pitches, which is still no because users not always the smortest but could be plausible for our backend

pybaseball/statcast_pitcher.py
```diff
@@ -25,3 +29,78 @@ def statcast_pitcher(start_dt: Optional[str] = None, end_dt: Optional[str] = Non
...
+def statcast_pitcher_percentile_ranks(year: int) -> pd.DataFrame:
+    url = f"https://baseballsavant.mlb.com/leaderboard/percentile-rankings?type=pitcher&year={year}&position=&team=&csv=true"
+    res = requests.get(url, timeout=None).content
+    data = pd.read_csv(io.StringIO(res.decode('utf-8')))
+    # URL returns a null player with player id 999999, which we want to drop
+    return data.loc[data.player_name.notna()].reset_index(drop=True)
+
+@cache.df_cache()
+def statcast_pitcher_spin_dir_comp(year: int, pitch_combo: str = "4 - Seamer / Changeup", minP: int = 100, pov: str = "Pit") -> pd.DataFrame:
+    pitches = ["4-Seamer", "Changeup", "Curveball", "Slider", "Sinker", "Cutter"]
```
schorrm, line 94: Fun fact: I checked the URLs a bit and this actually supports more pitches than this -- I think it supports the full range of the Statcast pitches. Splitter doesn't appear in the frontend, but it works.

schorrm: The one exception is 'All', it doesn't support that code

pybaseball/statcast_pitcher.py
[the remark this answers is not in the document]
schorrm: Here's a concrete suggestion. The idea is that all the pitch codes stuff would just use this normalize call. I know it's a bit longwinded, but it does result in clearer error messages ('ALL' especially can cause some trippy stuff)
```python
pitch_codes = ["FF", "SIFT", "CH", "CUKC", "FC", "SL", "FS", "ALL"] # note: all doesn't work in words, we'll have some special handling
pitch_names = ["4-Seamer", "Sinker", "Changeup", "Curveball", "Cutter", "Slider", "Sinker"]
pitch_names_upper = [p.upper() for p in pitch_names]

# including all the codes to themselves makes this simpler later
name_to_code_map = dict(zip(pitch_codes + pitch_names_upper, pitch_codes + pitch_codes))
code_to_name_map = dict(zip(pitch_codes, pitch_names))

def _norm_pitch_code(pitch: str, to_word: bool = False) -> str:
    # everything goes to code first to keep the code simple, since dict lookups are cheap
    try:
        pitch = name_to_code_map.get(pitch.upper())
    except KeyError:
        raise ValueError(f'{pitch} is not a valid pitch!')
    if to_word:
        if pitch == 'ALL':
            raise ValueError("'All' cannot be used in this function")
        pitch = code_to_name_map[pitch]
    return pitch
```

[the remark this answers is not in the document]
schorrm: A potential alternative could be:
```python
def _norm_pitch_code(pitch: str, to_word: bool = False) -> str:
    normed = name_to_code_map.get(pitch.upper())
    normed = code_to_name_map.get(normed) if to_word else normed
    if normed is None:
        raise ValueError(f'{pitch} is not a valid pitch!') # this is funky for all though
    return normed
```
This is much shorter, but it leads to weird messaging on the corner case of 'all' for a function that needs names, which don't accept 'all'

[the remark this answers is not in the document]
schorrm: Another potential alternative could be:
```python
def _norm_pitch_code(pitch: str, to_word: bool = False) -> str:
    normed = name_to_code_map.get(pitch.upper())
    normed = code_to_name_map.get(normed) if to_word else normed
    if normed is None:
        if pitch.lower() is 'all':
            raise ValueError("'All' is not a valid pitch in this particular context!")
        raise ValueError(f'{pitch} is not a valid pitch!') # this is funky for all though
    return normed
```
This covers that edge case

pybaseball/statcast_pitcher.py
TheCleric: In this case, is all just simulated with something like `','.join(pitches)` ?

wfordh: I think because it's comparing two different pitches it wouldn't support "All"

pybaseball/statcast_pitcher.py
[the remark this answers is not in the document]
wfordh: I like that last one. More succinct but covers everything. Would we want `_norm_pitch_code()` and the maps to live in `statcast_pitcher` or `utils`?

schorrm:
LGTM, and the checkmark gods are appeased.

wfordh:
@schorrm SGTM, feel free to merge whenever

Issue #188, opened 2021-03-11
Issues with battings_stats_range and pitching_stats_range


baileymorton989:
I have been trying to utilize the  `batting_stats_range()` and `pitching_stat_range()` functions but only `batting_stats_range()` seems to be able to handle a single date and a date range, whereas `pitching_stat_range()` can only handle a range.

For example, if I run the following code:
    
    import pybaseball as pyb
    import pandas as pd 
    import datetime

    #range
    start_time = '2019-05-01'
    end_time = '2019-05-08
    
    #make date range
    date_range = pd.date_range(start=start_time, end = end_time,freq='D')
    dates = {datetime.datetime.strftime(date,'%Y-%m-%d') : date for date in date_range}

    #grab stats
    batter_dic = {date:pyb.batting_stats_range(date,) for date in dates.keys()}
    pitcher_dic = {date:pyb.pitching_stats_range(date,) for date in dates.keys()}

I receive the error :

```
Error: a date range needs to be specified
Traceback (most recent call last):

  File "C:\Users\bam3k\OneDrive\Documents\Sports Projects\baseballforecaster_v_2_3.py", line 841, in <module>
    seasons_dict_pitcher = {dic:pyb.pitching_stats_range(dic,) for dic in years.keys()}

  File "C:\Users\bam3k\OneDrive\Documents\Sports Projects\baseballforecaster_v_2_3.py", line 841, in <dictcomp>
    seasons_dict_pitcher = {dic:pyb.pitching_stats_range(dic,) for dic in years.keys()}

  File "C:\Users\bam3k\Anaconda3\lib\site-packages\pybaseball\cache\cache.py", line 54, in _cached
    result = func(*args, **kwargs)

  File "C:\Users\bam3k\Anaconda3\lib\site-packages\pybaseball\league_pitching_stats.py", line 56, in pitching_stats_range
    table = get_table(soup)

  File "C:\Users\bam3k\Anaconda3\lib\site-packages\pybaseball\league_pitching_stats.py", line 25, in get_table
    table = soup.find_all('table')[0]

AttributeError: 'NoneType' object has no attribute 'find_all'
```

Is there a reason for this limitation?

baileymorton989:
Will this be fixed with the newest release?

schorrm:
Not with this one.

schorrm:
Sadly, not yet. I hope we will be able to deal with it soon.

baileymorton989:
[...]

PR #197, opened 2021-04-06
league pitching range fix

bdilday:
Fixes bug in applying date range to `pitching_stats_range`. Closes https://github.com/jldbc/pybaseball/issues/188


schorrm:
Good catch. I think I will merge this independently of everything else having gotten broken again.

baileymorton989:
I am still having the same issue that I had in #188.

PR #198, opened 2021-04-06
Statcast Fielding & Running leaderboards

wfordh:
Pulling data from statcast's fielding and running leaderboards. Can change to fielder and runner if that makes more sense / is more consistent. Most of it was pretty straightforward except for `statcast_outs_above_average()` due to how the positions are handled - if there's a better way to do `norm_positions()`, I'm all ears. 

The [OAA leaderboard](https://baseballsavant.mlb.com/leaderboard/outs_above_average?type=Fielder&year=2020&team=&range=year&min=q&pos=of&roles=&viz=show) in particular has a lot of options and using `kwargs` so the user can supply more args might be a good next step after this.

There's also no catcher pop time data for 2020, which I found odd. Once all of this checks out I will start on the documentation and tests.

### DIFF
--- a/docs/statcast_fielding.md
+++ b/docs/statcast_fielding.md
@@ -16,7 +16,7 @@ from pybaseball import statcast_outs_above_average
 data = statcast_outs_above_average(2019, "all", 50)
 
 # Center fielders who qualified in 2019
-data = statcast_outs_above_average(2019, "CF")
+data = statcast_outs_above_average(2019,  pos = "CF")
 
 # Infielders with at least 100 fielding attempts in 2019
 data = statcast_outs_above_average(2019, pos = "IF", min_att = 100)
--- a/docs/statcast_fielding.md
+++ b/docs/statcast_fielding.md
@@ -2,13 +2,13 @@
 **Note:** Statcast data is liable to change unexpectedly due to the large number of observations. Please keep that in mind when pulling data.
 
 # Statcast Fielding Outs Above Average
-`statcast_outs_above_average(year: int, pos: str, min_att: Union[int, str] = "q")`
+`statcast_outs_above_average(year: int, pos: Union[int, str], min_att: Union[int, str] = "q")`
 
 This function retrieves outs above average (OAA) for the given year, position, and attempts. OAA is a Statcast metric based on the "cumulative effect of all individual plays a fielder has been credited or debited with, making it a range-based metric of fielding skill that accounts for the number of plays made and the difficulty of them".
 
 ## Arguments
 `year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
-`pos:` The position you are interested in. Valid positions include "all", "IF", "OF", and position names or abbreviations. Pitchers and catchers are not included.
+`pos:` The position you are interested in. Valid positions include "all", "IF", "OF", and position names, numbers, or abbreviations. Position numbers may be entered as integers or strings, e.g. 6 or "6" for shortstops. Pitchers and catchers are not included.
 `min_att:` The minimum number of fielding attempts for the player to be included in the result. Statcast's default is players, which is 1 fielding attempt per game played for 2B, SS, 3B, and OF and 1 fielding attempt per every other game played for 1B.
 
 ## Examples of Valid Queries
@@ -21,6 +21,8 @@ data = statcast_outs_above_average(2019, "all", 50)
 # Center fielders who qualified in 2019
 data = statcast_outs_above_average(2019,  pos = "cf")
 
+# Shortstops who qualified in 2019
+data = statcast_outs_above_average(2019, pos = 6)
 # Infielders with at least 100 fielding attempts in 2019
 data = statcast_outs_above_average(2019, pos = "IF", min_att = 100)
 ```
--- a/docs/statcast_fielding.md
+++ b/docs/statcast_fielding.md
@@ -1,0 +1,57 @@
+# Statcast Fielding Outs Above Average
+`statcast_outs_above_average(year: int, pos: str, min_att: Union[int, str] = "q")`
+
+This function retrieves outs above average (OAA) for the given year, position, and attempts. OAA is a Statcast metric based on the "cumulative effect of all individual plays a fielder has been credited or debited with, making it a range-based metric of fielding skill that accounts for the number of plays made and the difficulty of them".
+
+## Arguments
+
+## Examples of Valid Queries
+
+
+# Statcast Fielding Outfield Directional OAA
+`statcast_outfield_directional_oaa(year: int, min_opp: Union[int, str] = "q")`
+
+<!-- desc -->
+
+## Arguments
+
+## Examples of Valid Queries
+
+
+# Statcast Fielding Outfield Catch Probability
+`statcast_outfield_catch_proba(year: int, min_opp: Union[int, str] = "q")`
+
+<!-- desc -->
+
+## Arguments
+
+## Examples of Valid Queries
+
+# Statcast Fielding Outfielder Jump
+`statcast_outfielder_jump(year: int, min_att: Union[int, str] = "q")`
+
+<!-- desc -->
+
+## Arguments
+
+## Examples of Valid Queries
+
+# Statcast Fielding Catcher Poptime
+`statcast_catcher_poptime(year: int, min_2b_att: int, min_3b_att: int)`
+
+<!-- desc -->
+
+## Arguments
+
+## Examples of Valid Queries
+
+
+# Statcast Fielding Catcher Framing
+`statcast_catcher_framing(year: int, min_called_p: Union[int, str] = "q")`
+
+<!-- desc -->
+
+## Arguments
+
+## Examples of Valid Queries
+
--- a/docs/statcast_fielding.md
+++ b/docs/statcast_fielding.md
@@ -48,7 +48,7 @@ data = statcast_outfield_directional_oaa(2019, 200)
 ```
 
 # Statcast Fielding Outfield Catch Probability
-`statcast_outfield_catch_proba(year: int, min_opp: Union[int, str] = "q")`
+`statcast_outfield_catch_prob(year: int, min_opp: Union[int, str] = "q")`
 
 This function retrieves aggregated data for outfielder performance on fielding attempt types, binned into five star categories, for the given year and number of opportunities.
 
@@ -58,13 +58,13 @@ This function retrieves aggregated data for outfielder performance on fielding a
 
 ## Examples of Valid Queries
 ```python
-from pybaseball import statcast_outfield_catch_proba
+from pybaseball import statcast_outfield_catch_prob
 
 # All qualified outfielders from 2019
-data = statcast_outfield_catch_proba(2019)
+data = statcast_outfield_catch_prob(2019)
 
 # All outfielders with at least 200 attempts in 2019
-data = statcast_outfield_catch_proba(2019, 200)
+data = statcast_outfield_catch_prob(2019, 200)
 ```
 
 # Statcast Fielding Outfielder Jump
--- a/docs/statcast_fielding.md
+++ b/docs/statcast_fielding.md
@@ -6,7 +6,7 @@ This function retrieves outs above average (OAA) for the given year, position, a
 ## Arguments
 `year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
 `pos:` The position you are interested in. Valid positions include "all", "IF", "OF", and position names or abbreviations. Pitchers and catchers are not included.
-`min_att:` The minimum number of fielding attempts for the player to be included in the result. The default is players, which is 1 fielding attempt per game played for 2B, SS, 3B, and OF and 1 fielding attempt per every other game played for 1B.
+`min_att:` The minimum number of fielding attempts for the player to be included in the result. Statcast's default is players, which is 1 fielding attempt per game played for 2B, SS, 3B, and OF and 1 fielding attempt per every other game played for 1B.
 
 ## Examples of Valid Queries
 ```python
@@ -29,7 +29,7 @@ This function retrieves outfielders' directional OAA data for the given year and
 
 ## Arguments
 `year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
-`min_opp:` The minimum number of opportunities for the player to be included in the result. The default is players with at least 1 fielding attempt per game.
+`min_opp:` The minimum number of opportunities for the player to be included in the result. Statcast's default is players with at least 1 fielding attempt per game.
 
 ## Examples of Valid Queries
 ```python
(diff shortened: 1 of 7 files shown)


pybaseball/statcast_fielding.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Optional, Union
+
+import pandas as pd
+import requests
+
+from . import cache
+from .utils import norm_positions
+
+@cache.df_cache()
+def statcast_outs_above_average(year: int, posn: str, min_att: Union[int, str] = "q") -> pd.DataFrame:
```
schorrm, line 11: Isn't 'pos' a more typical abbreviation? (would apply for usages throughout)

pybaseball/statcast_running.py
```diff
@@ -0,0 +1,24 @@
...
+
+@cache.df_cache()
+def statcast_running_splits(year: int, min_opp: int = 5, split_type: str = "raw") -> pd.DataFrame:
+	splits = ["raw", "percent"]
+	if split_type not in splits:
+		raise ValueError(f"Not a valid split_type. Must be one of {', '.join(splits)}.")
+	url = f"https://baseballsavant.mlb.com/running_splits?type={split_type}&bats=&year={year}&position=&team=&min={min_opp}&csv=true"
+	res = requests.get(url, timeout=None).content
+	data = pd.read_csv(io.StringIO(res.decode('utf-8')))
+	return data
```
schorrm, line 24: bonk

pybaseball/statcast_running.py
```diff
@@ -0,0 +1,24 @@
...
+	url = f"https://baseballsavant.mlb.com/leaderboard/sprint_speed?year={year}&position=&team=&min={min_opp}&csv=true"
+	res = requests.get(url, timeout=None).content
+	data = pd.read_csv(io.StringIO(res.decode('utf-8')))
+	return data
+
+@cache.df_cache()
+def statcast_running_splits(year: int, min_opp: int = 5, split_type: str = "raw") -> pd.DataFrame:
+	splits = ["raw", "percent"]
+	if split_type not in splits:
+		raise ValueError(f"Not a valid split_type. Must be one of {', '.join(splits)}.")
```
schorrm, line 20: I would suggest that for just two options change this to a bool

pybaseball/utils.py
```diff
@@ -281,10 +290,23 @@ def flag_imputed_data(statcast_df: pd.DataFrame) -> pd.DataFrame:
...
+    normed = posn_name_to_code_map.get(posn.upper())
+    normed = posn_code_to_name_map.get(normed) if to_word else normed
+    if to_number:
+        if normed not in ["IF", "OF"]:
+            normed = posn_code_to_numbers_map.get(normed)
+        if posn.lower() == "all":
+            normed = ""
+    if normed is None:
+        raise ValueError(f'{posn} is not a valid position!')
+    return normed.lower()
```
schorrm, line 311: This whole thing has type problems (and IF / OF don't have numeric norms here?)

schorrm (changes requested):

pybaseball/statcast_running.py
schorrm: (EOF)

pybaseball/statcast_running.py
wfordh: Ah yeah will update that

pybaseball/statcast_fielding.py
wfordh: Yeah I guess it is...I think I went with `posn` to differentiate from `positive`, but will change that

pybaseball/utils.py
wfordh: Yeah this was tough to figure out. The Statcast url uses numbered positions for when you pick a single one, but then for `IF / OF`, the url requires lowercase `if / of` and for `all`, the url just uses an empty string. Not the best for what we're trying to do

schorrm: Yeah but you're potentially calling lower on an int

wfordh: Technically it isn't since I'm casting the ints to strings when created the map earlier in the code, so actually I didn't need to update the return types like I did - it'll always return a string. I can add some in-line comments to make that clearer or move the `lower()` call to somewhere that makes more sense

schorrm:
Looks like a good initial draft here -- ideally, I would prefer to bind the parameters for the OAA leaderboard to the func call, rather than just dangle in the thingy with kwargs.

schorrm:
This is a design decision though -- other than this point, the rest looks ready to move on to testing + documentation, this decision should be made and then

wfordh:
@schorrm LMK how it looks when you get a chance

tests/integration/pybaseball/test_statcast_fielding.py
```diff
@@ -18,7 +18,7 @@ def test_statcast_outs_above_average() -> None:
 	assert not result.empty
 
 	assert len(result.columns) == 17
-	assert len(result) == 184
+	assert len(result) == 182
```
schorrm, line 21: Relax this assert
Read the article: "You'll never believe these 42 HIGHLY EFFECTIVE ways Tom Tango found to mess with @TheCleric"

tests/integration/pybaseball/test_statcast_fielding.py
```diff
@@ -39,7 +39,7 @@ def test_statcast_outfield_catch_proba() -> None:
 	assert not result.empty
 
 	assert len(result.columns) == 19
-	assert len(result) == 228
+	assert len(result) == 224
```
schorrm, line 42: Relax this assert.
After all, what is 224 if is not just a positive number 😆 / 😭?

tests/integration/pybaseball/test_statcast_running.py
```diff
@@ -17,14 +16,13 @@ def test_statcast_sprint_speed() -> None:
...
 	min_opp = 5
 	raw_splits = True
 	result: pd.DataFrame = statcast_running_splits(2019, min_opp, raw_splits)
 
 	assert result is not None
-	assert not result.empt
+	assert not result.empty
 
 	assert len(result.columns) == 27
 	assert len(result) == 546
```
schorrm, line 28: Relax this assert too

docs/statcast_fielding.md
```diff
@@ -0,0 +1,123 @@
+# Statcast Fielding Outs Above Average
+`statcast_outs_above_average(year: int, pos: str, min_att: Union[int, str] = "q")`
+
+This function retrieves outs above average (OAA) for the given year, position, and attempts. OAA is a Statcast metric based on the "cumulative effect of all individual plays a fielder has been credited or debited with, making it a range-based metric of fielding skill that accounts for the number of plays made and the difficulty of them".
+
+## Arguments
+`year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
+`pos:` The position you are interested in. Valid positions include "all", "IF", "OF", and position names or abbreviations. Pitchers and catchers are not included.
```
schorrm, line 8: Somewhat unrelated parenthetical: who the hell would rather use the position names???

docs/statcast_fielding.md
```diff
@@ -0,0 +1,123 @@
...
+from pybaseball import statcast_outs_above_average
+
+# All fielders with at least 50 fielding attempts in 2019
+data = statcast_outs_above_average(2019, "all", 50)
+
+# Center fielders who qualified in 2019
+data = statcast_outs_above_average(2019,  pos = "CF")
+
+# Infielders with at least 100 fielding attempts in 2019
+data = statcast_outs_above_average(2019, pos = "IF", min_att = 100)
```
schorrm, line 22: Should we note explicitly or by example that this isn't case-sensitive?

pybaseball/statcast_fielding.py
```diff
@@ -0,0 +1,59 @@
...
+
+@cache.df_cache()
+def statcast_outfielder_jump(year: int, min_att: Union[int, str] = "q") -> pd.DataFrame:
+	url = f"https://baseballsavant.mlb.com/leaderboard/outfield_jump?year={year}&min={min_att}&csv=true"
+	res = requests.get(url, timeout=None).content
+	data = pd.read_csv(io.StringIO(res.decode('utf-8')))
+	return data
+
+@cache.df_cache()
+def statcast_catcher_poptime(year: int, min_2b_att: int, min_3b_att: int) -> pd.DataFrame:
```
schorrm, line 45: Is there no 'q' here? The signature is different

docs/statcast_fielding.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+# Statcast Fielding Outfield Directional OAA
+`statcast_outfield_directional_oaa(year: int, min_opp: Union[int, str] = "q")`
+
+This function retrieves outfielders' directional OAA data for the given year and number of opportunities. The directions are Back Left, Back, Back Right, In Left, In, and In Right.
+
+## Arguments
+`year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
+`min_opp:` The minimum number of opportunities for the player to be included in the result. The default is players with at least 1 fielding attempt per game.
```
schorrm, line 32: This default is provided by Statcast, so we should probably say:
`Statcast's default is players with at least 1 fielding attempt per game (so this is liable to change unexpectedly).`
(@TheCleric)

docs/statcast_fielding.md
```diff
@@ -0,0 +1,123 @@
...
+
+# All qualified outfielders from 2019
+data = statcast_outfield_directional_oaa(2019)
+
+# All outfielders with at least 200 attempts in 2019
+data = statcast_outfield_directional_oaa(2019, 200)
+```
+
+# Statcast Fielding Outfield Catch Probability
+`statcast_outfield_catch_proba(year: int, min_opp: Union[int, str] = "q")`
```
schorrm, line 46: Wouldn't 'prob' be a more normal condensation (and hence `statcast_outfield_catch_prob`)?

docs/statcast_fielding.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+# Statcast Fielding Catcher Framing
+`statcast_catcher_framing(year: int, min_called_p: Union[int, str] = "q")`
+
+This function retrieves the catcher's framing results for the given year and minimum called pitches. It uses eight zones around the strike zone (aka "shadow zone") and gives the percentage of time the catcher gets the strike called in each zone.
+
+## Arguments
+`year:` The year for which you wish to retrieve batted ball against data. Format: YYYY.
+`min_called_p:` The minimum number of called pitches for the catcher in the shadow zone. The default is players with at least 6 called pitches in the shadow zone per team game.
```
schorrm, line 112: Given how this recurs, maybe just say "Statcast's default" instead of "The default"

pybaseball/statcast_fielding.py
```diff
@@ -0,0 +1,59 @@
...
+import pandas as pd
+import requests
+
+from . import cache
+from .utils import norm_positions
+
+@cache.df_cache()
+def statcast_outs_above_average(year: int, pos: str, min_att: Union[int, str] = "q") -> pd.DataFrame:
+	# a lot of options here. need to handle all the positions. range and split years?
+	# can use startYear and endYear for multi year
```
schorrm, line 13: What's the deal with this comment?

pybaseball/statcast_fielding.py
```diff
@@ -0,0 +1,59 @@
...
+	pos = norm_positions(pos)
+	# catcher is not included in this leaderboard
+	if pos == "2":
+		raise ValueError("'C' not a valid position in this particular context!")
```
schorrm, line 17: Would `"This particular leaderboard does not include catchers!"` be clearer?

pybaseball/statcast_fielding.py
```diff
@@ -0,0 +1,59 @@
...
+	url = f"https://baseballsavant.mlb.com/leaderboard/outs_above_average?type=Fielder&year={year}&team=&range=year&min={min_att}&pos={pos}&roles=&viz=show&csv=true"
```
schorrm, line 18: Is `&viz=show` of any consequence when `csv=true`?

schorrm (changes requested):

schorrm:
Sorry for the nutso review, but this is big. Great work, even if the review is only about quibbling. Sorry.

docs/statcast_fielding.md
wfordh: Not sure, but I included them in case anyone does. I think motivated by how we handled the pitch names. Could maybe switch it up to accept position numbers?

docs/statcast_fielding.md
wfordh: I think by example is fine. I'll make one of the examples here lowercase

pybaseball/statcast_fielding.py
wfordh: There is not, they just had it set to 5 and 0. I'll add those as the defaults.

docs/statcast_fielding.md
wfordh: Should we just put that disclaimer at the top of the docs and say it applies for all of them?

docs/statcast_fielding.md
wfordh: Yeah it could be. I'm used to scikit-learn's usage where they have `predict_proba()` on things like the [Random Forest Classifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier.predict_proba)

pybaseball/statcast_fielding.py
wfordh: oops old notes. will remove

pybaseball/statcast_fielding.py
wfordh: Nope, works fine without it or with it set to `viz=hide`

wfordh:
No worries at all! Better to cover all of it now

docs/statcast_fielding.md
schorrm: No need to remove, maybe adding numbers could be good 🤷

docs/statcast_fielding.md
schorrm: Also good

pybaseball/statcast_fielding.py
```diff
@@ -42,7 +40,7 @@ def statcast_outfielder_jump(year: int, min_att: Union[int, str] = "q") -> pd.Da
 	return data
 
 @cache.df_cache()
-def statcast_catcher_poptime(year: int, min_2b_att: int, min_3b_att: int) -> pd.DataFrame:
+def statcast_catcher_poptime(year: int, min_2b_att: int = 5, min_3b_att: int = 0) -> pd.DataFrame:
```
schorrm, line 43: So now it's not a statcast default?

wfordh: It is what Statcast defaults to, they just don't have a `qualified` option for this leaderboard

wfordh:
Oops, had an issue with tabs / spaces in `utils.py` that caused a big "change", but only material change there is at the beginning of `norm_positions()`

PR #201, opened 2021-04-07
fix test failures

schorrm:

PR #200, opened 2021-04-06
get a logger instead of globally config logging

Maradonna90:
fixes #199

PR #202, opened 2021-04-07
Various test fixes

TheCleric:


schorrm:
Rejecting this in favor of #200 and #201 separately.

PR #205, opened 2021-04-12
Cache tests updates

bdilday:
This attempts to stop the sporadic failures of the cache tests. It mocks the `safe_jsonify` method, which is called by the `save` method of a cache record. This prevents any files from being written during the cache testing. This will hopefully address the failures of `test_call_cache_ignores_expired`, `test_call_cache_gets_uncached_data`, `test_call_cache_load_fails_silently`.

Because there are now no files to load it also has to mock the json load method in the `test_call_cache_save_fails_silently` test.


schorrm:
LGTM

PR #206, opened 2021-04-25
Update split_stats.py

Latonis:
Baseball-reference html has changed for split stats.

 {"class": "overthrow table_container"} is now  {"class": "table_container"}


wfordh:
LGTM cc @schorrm

PR #207, opened 2021-04-25
Make parallel statcast requests optional

reddigari:
Adds a `parallel` parameter to `pybaseball.statcast()`. If `False`, requests will be made sequentially rather than through a `ProcessPoolExecutor`.

The motivation for this is that I was trying to call the function in an AWS lambda function, where all multiprocessing is disabled. This might be a somewhat niche use case, so no worries if you don't want it. I didn't look into covering it in the tests, but I'm happy to.


schorrm:
Seems fair enough

schorrm:
Seems fair enough

schorrm:
Looks good, can you just update the docs to match?

schorrm:
LGTM

PR #222, opened 2021-07-14
Scrape MLB IDs from Baseball-Reference

marek-slipski:
## Context
Baseball-Reference player stat pages ([like this](https://www.baseball-reference.com/leagues/daily.fcgi?user_team=&bust_cache=&type=b&lastndays=7&dates=fromandto&fromandto=20190410.20190411&level=mlb&franch=&stat=&stat_value=0)) have player MLB IDs embedded in links. Scraping these IDs would be useful as `batting_stats_range` and `pitching_stats_range` return a Dataframe with Baseball-Reference player names as the only identifier but these are difficult to join on. This would grab the IDs from the links and add an `mlbID` column to the batting and pitching tables.

Example output of `pitching_stats_range`:
image


schorrm:
We can do this if you update the unit tests to match

marek-slipski:
I think that should do it, but let me know if I'm missing something (I'm new at this).

schorrm:
whoops, missed. can you merge the upstream testing changes back in and push and see if it passes the CI?

Issue #218, opened 2021-06-28
batting_stats_range breaks when parsing through the date '2021-06-25'


dk-sa1:
When using the batting_stats_range function, there is an issue when parsing through 6/25/2021. 

When it breaks, you receive corrupt data. One such piece being the player José Abreu appearing as "JosÃ© Abreu". As well as only receiving a couple rows of data ( As opposed to several hundred for a typical day of data). 

Below are some code blocks that work and do not work. 

Works:
data = batting_stats_range("2021-06-25", "2021-06-27")
data = batting_stats_range("2021-06-25", "2021-06-25")

Does NOT Work:
data = batting_stats_range("2021-06-24", "2021-06-27")
data = batting_stats_range("2021-06-24", "2021-06-25")

For some reason, you can start on 6/25 with no issues. But you cannot parse over, nor end on 6/25 without receiving corrupt data.

dk-sa1:
The issue arises for the same date in 2019.

jmcous:
I'm having other problems with batting_stats_range, I don't know if this is related but:
`from IPython.display import display

a=batting_stats_range('2021-05-25','2021-06-23')
b=pitching_stats_range('2021-05-25','2021-06-23')
print(a.shape)
print(b.shape)

display(a)
display(b)`

The batting stats range query only returns 21 rows, just players that start with 'A'.
![image](https://user-images.githubusercontent.com/29820512/123888688-ac1e2a80-d919-11eb-99b5-9bdf4ae4b2bc.png)

bdilday:
it looks like in the cases that the data gets truncated, beautiful soup can't use utf-8 so falls back on a different encoding, e.g., 

```
>>> from pybaseball.league_batting_stats import batting_stats_range, get_soup
>>> start_dt = end_dt = "2021-05-01"
>>> data = batting_stats_range(start_dt, end_dt)
>>> len(data)
334
>>> soup = get_soup(start_dt, end_dt)
>>> soup.original_encoding
'utf-8'
>>> 
>>> start_dt = end_dt = "2021-05-02"
>>> data = batting_stats_range(start_dt, end_dt)
>>> len(data)
10
>>> soup = get_soup(start_dt, end_dt)
>>> soup.original_encoding
'Windows-1252'
>>> 
```

[...]

PR #223, opened 2021-07-17
bugfix for truncated batting range data

bdilday:
Forces a utf-8 encoding for baseball reference data in `batting_stats_range` and `pitching_stats_range` before applying beautiful soup parsing. This should avoid confusion in the encoding due to inconsistency between the table and the footer of the page.
Closes https://github.com/jldbc/pybaseball/issues/218


schorrm:
LGTM

Issue #225, opened 2021-07-27
statcast_sprint_speed Import Error


jgamblin:
I was attempting to look at the sprint speeds for this year using [this example](https://github.com/jldbc/pybaseball/blob/master/docs/statcast_running.md ) and end up getting this error:

 ```ImportError: cannot import name 'statcast_sprint_speed' from 'pybaseball' (/usr/local/lib/python3.9/site-packages/pybaseball/__init__.py)```

[My Test Jupyter Notebook is here. ](https://gist.github.com/jgamblin/0491f84a0d7fc6a50e394b3ed4692c89)

![Screen Shot 2021-07-27 at 7 50 33 AM](https://user-images.githubusercontent.com/8428124/127156446-70cfcdfb-d1e0-40fc-85b0-87c62c18b426.png)

schorrm:
Whoopsy, idk what happened there.
For the time being, you can use:
`from pybaseball.statcast_running import statcast_sprint_speed`
@wfordh what should be happening here? Is the doc wrong or is the code wrong?

wfordh:
Hmm I'll look into it. I know when I was testing the other statcast pieces the imports were a little different than anticipated. Without looking, either what's in the example or what @schorrm suggested should work.

schorrm:
Fixed in #228

PR #228, opened 2021-08-01
Fix import statements in statcast docs

wfordh:
Fixing issues with documentation for statcast docs brought up by #225

### DIFF
--- a/docs/statcast_batter.md
+++ b/docs/statcast_batter.md
@@ -41,7 +41,7 @@ This function retrieves batted ball data for all batters in a given year.
 
 ## Examples of Valid Queries
 ```python
-from pybaseball.statcast_batter import statcast_batter_exitvelo_barrels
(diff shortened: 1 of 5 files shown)


bdilday:
I think it'd be a good idea to expose these at the top level, example, in `pybaseball/__init__.py`
```
from .statcast_batter import (
    statcast_batter,
    statcast_batter_exitvelo_barrels,
    statcast_batter_expected_stats,
    statcast_batter_percentile_ranks,
    statcast_batter_pitch_arsenal,
)
```

wfordh:
Ah so _that's_ why I could import `statcast_batter()` directly and couldn't with these. I can add the functions into `pybaseball/__init__.py` like that for all of them and then revert back to the original documentation

schorrm:
LGTM, @bdilday all good?

PR #232, opened 2021-08-11
Add a matrix to test on all python versions

TheCleric:
Testing out a matrix that will test changes against all current python versions.

@schorrm I think this could help catch things like what I found in #231 which was a TypeError that seems to happen in 3.7+, but not previously caught in our 3.6 tests (which you can see are causing failures below in 3.7 and 3.8).

As well we may be able to easily add 3.10 when that's ready to see how our code base will fair.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -9,14 +9,21 @@ on:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        python-version:
+        - 3.6
+        - 3.7
+        - 3.8
+        - 3.9
     name: Install dependencies and test
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


schorrm:
I would prefer to not commit to supporting so many in the tests, it also has time costs

TheCleric:
> I would prefer to not commit to supporting so many in the tests, it also has time costs

They run in parallel, so the time cost should be minimal.

schorrm:
red exs scary

TheCleric:
> red exs scary

They'll get fixed after the other PR gets merged (and master gets merged in here)

schorrm:
Other merged, can you merge back and appease the CI?

TheCleric:
@schorrm all tests passing now in case you didn't see it.

PR #241, opened 2021-10-18
Created a new function to retrieve box scores from baseball reference…

demilio76:
…. Quick example:

datetime_object = datetime.strptime('May 05 2021', '%b %d %Y')
    visitor_batting_df, home_batting_df, visitor_pitching_df, home_pitching_df \
        = box_score('OAK', datetime_object, 0)
    print(f"{visitor_pitching_df.loc[0, 'Pitching']} vs {home_pitching_df.loc[0, 'Pitching']}")

### DIFF
--- a/pybaseball/box_score.py
+++ b/pybaseball/box_score.py
@@ -1,0 +1,156 @@
+import datetime
+import re
+
+import pandas as pd
+import requests
+from bs4 import BeautifulSoup
+
+
+def post_process_batting(batter_df, html_table):
+    """
+       Retrieves the bbref player_id from a html header attribute, determines the starters/positions,
+       and cleans up the name
+    """
+    row_count = 0
+    table_soup = BeautifulSoup(html_table, "lxml")
+    for row in table_soup.findAll('tr'):
+        if 0 < row_count < len(batter_df.index):  # skip the first row of the table plus any trailing rows
+            header = row.find('th')
+
+            if 'iz' not in header['class']:
+                # first get the player_id for each batter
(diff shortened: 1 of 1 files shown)


schorrm:
@TheCleric @bdilday if either of you can take a look?

bdilday:
> I had to parse through comments to get the data I wanted due to how bbref sets up their boxscore pages. Alternatively, I have a version that uses Selenium and a ChromeDriver which works a little cleaner (tables aren't in comments post-page load) but for now am submitting this version to avoid a new dependency

I think I'd rather have the cleaner, selenium, version. It doesn't seem like a crazy dependency for a library who's job is largely to scrape the web.

@schorrm @TheCleric any thoughts?

TheCleric:
> > I had to parse through comments to get the data I wanted due to how bbref sets up their boxscore pages. Alternatively, I have a version that uses Selenium and a ChromeDriver which works a little cleaner (tables aren't in comments post-page load) but for now am submitting this version to avoid a new dependency
> 
> I think I'd rather have the cleaner, selenium, version. It doesn't seem like a crazy dependency for a library who's job is largely to scrape the web.
> 
> @schorrm @TheCleric any thoughts?

I'm not a fan of selenium for this since it doesn't need anything like JavaScript. As it is we've done similar things to this with just the xpath parser which can be used to parse into HTML comments.

BrayanMnz:
This has been opened through a year - are we merging this into the project or not?
@schorrm @tjburch

Issue #242, opened 2021-11-15
team_results.py `FutureWarning`


notBillJames:
On lines 72, and 85 no keywords are passed for arguments into the `DataFrame.drop() function`. The following `FutureWarning` is returned:

```
pybaseball/pybaseball/team_results.py:72: FutureWarning: In a future version of pandas all arguments of DataFrame.drop except for the argument 'labels' will be keyword-only
  df = df.drop('',1) #not a useful column
```

It would be good to change this to avoid deprecation.

tjburch:
Thanks for noting this, just made a quick PR to resolve.

schorrm:
Closed by #243

PR #243, opened 2021-11-15
Add axis keyword to df.drop call

tjburch:
Issue #242 notes that drop now throws a `FutureWarning` for non-keyword args beyond labels. Adding the `axis` keyword so this warning doesn't persist.

### DIFF
--- a/pybaseball/team_results.py
+++ b/pybaseball/team_results.py
@@ -69,7 +69,7 @@ def get_table(soup: BeautifulSoup, team: str) -> pd.DataFrame:
     df = pd.DataFrame(data)
     df = df.rename(columns=df.iloc[0])
     df = df.reindex(df.index.drop(0))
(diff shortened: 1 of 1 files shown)


pybaseball/team_results.py
```diff
@@ -69,7 +69,7 @@ def get_table(soup: BeautifulSoup, team: str) -> pd.DataFrame:
     df = pd.DataFrame(data)
     df = df.rename(columns=df.iloc[0])
     df = df.reindex(df.index.drop(0))
-    df = df.drop('',1) #not a useful column
+    df = df.drop('',axis=1) #not a useful column
```
schorrm, line 72: LGTM, but nit: space?

tjburch: Lol leave it to me to mess up a 5 character commit.

PR #252, opened 2022-02-27
Update test configuration + tests

TheCleric:


schorrm:
Don't love all the >0 but let's stop the bleeding

PR #256, opened 2022-03-18
Bug fix for `statcast_batter`

### LINKED ISSUE #255: [Bug] statcast_batter must be passed an end datetime
`statcast_batter` is designed to have optional start/end times, and the logic for handling them is in the call to `sanitize_input`. However, the code does not reassign the inputs after the call to `sanitize_input`, and the assertions end up failing.

akern40:
`statcast_batter` asserts that its parameters are not `None`, but does not reassign the parameters after the call to `sanitize_input`, leading to failure when `end_dt` is left as `None`. This commit reassigns the inputs from `sanitize_input`. `player_id` is not reassigned to avoid typing errors, since `sanitize_input` turns it into a string but `split_request` still expects an integer. Closes #255


schorrm:
Looks great, can you add a test case to catch this?

akern40:
@schorrm just added a commit that tests whether optional start and end dates autofill as expected.

Issue #267, opened 2022-05-14
pybaseball.pitching_stats() returns batting stats, not pitching stats


armstjc:
When running the following code:

```
import pybaseball
import pandas as pd

def getFanGraphsBattingStats(start=2020, end=2021):
    for i in range(start,end+1):
        data = pybaseball.batting_stats(i)
        data.to_csv(f'Data/FanGraphs/Batting/{i}_fangraphs_batting.csv',index=False)
        print(data)

def getFanGraphsPitchingStats(start=2020, end=2021):
    for i in range(start,end+1):
        data = pybaseball.pitching_stats(i)
        data.to_csv(f'Data/FanGraphs/Pitching/{i}_fangraphs_pitching.csv',index=False)
        print(data)


if __name__ == "__main__":
    pybaseball.cache.enable()
    getFanGraphsBattingStats()
    getFanGraphsPitchingStats()
```
It does not get the pitching stats as intended, as shown below:
```

      IDfg  Season                Name   Team  ...   CSW%    xBA   xSLG  xwOBA
18   19709    2020  Fernando Tatis Jr.    SDP  ...  0.269  0.297  0.614  0.419
1     5361    2020     Freddie Freeman    ATL  ...  0.191  0.341  0.660  0.464
4    13510    2020        Jose Ramirez    CLE  ...  0.229  0.263  0.505  0.371
9    15676    2020          Jose Abreu    CHW  ...  0.293  0.299  0.587  0.398
20   13611    2020        Mookie Betts    LAD  ...  0.272  0.281  0.481  0.359
..     ...     ...                 ...    ...  ...    ...    ...    ...    ...
130  13145    2020           Josh Bell    PIT  ...  0.284  0.228  0.381  0.297
139   6153    2020     Eduardo Escobar    ARI  ...  0.260  0.261  0.394  0.305
117   3892    2020        Josh Reddick    HOU  ...  0.260  0.245  0.358  0.300
128   6184    2020       J.D. Martinez    BOS  ...  0.263  0.229  0.444  0.316
136  10071    2020     Jonathan Villar  - - -  ...  0.266  0.211  0.281  0.256

[142 rows x 319 columns]
      IDfg  Season                Name   Team  ...   CSW%    xBA   xSLG  xwOBA
3    19709    2021  Fernando Tatis Jr.    SDP  ...  0.270  0.279  0.618  0.406
1    20123    2021           Juan Soto    WSN  ...  0.263  0.304  0.544  0.430
[...]

Issue #221, opened 2021-07-06
Pitching / Batting Stats Mix-up


zach-lewis:
I've been having issues with `batting_stats` and `pitching_stats` returning unexpected data based on date parameters. Originally, `pybaseball.pitching_stats(2021)` was returning what would be expected from  `pybaseball.batting_stats(2021)` (batting stats). This could originally be resolved by instead utilizing `pybaseball.pitching_stats(2020, 2021)` and filtering out 2020. 

However, now `pitching_stats` is behaving normally, but `batting_stats` is returning pitching stats for any years beyond 2018. Any ideas why this is happening?

Doddy-codes:
Bump... I am having a similar issue... 

Batting stats and pitching stats are both returning the same frame

df = batting_stats(2017, ind=1)  # get batting stats for season
pitch = pitching_stats(2017, ind=1)
pitch.equals(df)  # outputs true

tjburch:
Which version of pybaseball are you using? I wasn't able to reproduce your example with 2.2.1 which is what pip grabs for me.

```
In [1]: from importlib.metadata import version

In [2]: import pybaseball

In [3]: version('pybaseball')
Out[3]: '2.2.1'

In [4]: batting = pybaseball.batting_stats(2017, ind=1)

In [5]: pitching = pybaseball.pitching_stats(2017, ind=1)

In [6]: batting.equals(pitching)
Out[6]: False

In [7]: batting.head()
Out[7]:
     IDfg  Season               Name Team  Age    G   AB   PA    H   1B  2B  3B  HR    R  RBI   BB  ...  Soft%+  Med%+  Hard%+    EV    LA  Barrels  Barrel%  maxEV  HardHit  HardHit%  Events  CStr%   CSW%    xBA   xSLG  xwOBA
1   15640    2017        Aaron Judge  NYY   25  155  542  678  154   75  24   3  52  128  114  127  ...      61     88     140  94.9  15.8       84    0.249  121.1      186     0.550     338  0.157  0.290  0.281  0.651  0.441
6    5417    2017        Jose Altuve  HOU   27  153  590  662  204  137  39   4  24  112   81   58  ...     103    107      87  86.1   9.7       29    0.057  109.1      143     0.280     511  0.161  0.233  0.271  0.445  0.340
[...]

PR #269, opened 2022-05-20
Use objects qualified name in caching 

bdilday:
This adds a CI stage to illustrate the problem mentioned here https://github.com/jldbc/pybaseball/issues/267 https://github.com/jldbc/pybaseball/issues/221
namely that when the cache is enabled, the call to get the fangraphs batting and pitching stats return the same dataframe.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -31,6 +31,8 @@ jobs:
       - name: Timing test
         timeout-minutes: 2
         run: python -m scripts.statcast_timing
+      - name: Cache test
+        run: make validate-cache
       - name: Run MyPy
         run: make mypy ONLY_MODIFIED=0
         continue-on-error: true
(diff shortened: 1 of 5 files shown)


bdilday:
Updated the MR to include a proposed fix for the issue of batting and pitching stats returning the same dataframe. 

* moves the cache decorator for the `fetch` method from the abstract base class `FangraphsDataTable` to the concrete classes `FangraphsBattingStatsTable`, ` FangraphsFieldingStatsTable`, `FangraphsPitchingStatsTable`

this means the fully qualified name will include the concrete class name (e.g. `FangraphsBattingStatsTable.fetch`) not the base class (e.g.  `FangraphsDataTable.fetch`

* when hashing the function name for caching, uses the fully qualified name (if it doesn't contain `<locals>`) instead of just the name. this means it will be e.g. `FangraphsBattingStatsTable.fetch` instead of `fetch`

PR #274, opened 2022-06-08
Add team option for statcast batter & pitcher

nickbronson:
Add split_request_team() to utils.py, similar to split_request() but changed the player_id parameter to the 3 character team id.  Added this to not break any existing functionality with split_request() in statcast_batter.py and statcast_pitcher.py.  

Added statcast_batter_team() to statcast_batter.py to pull in all batters for a specified team during the time range.  Similar to statcast_batter() but changed the player_id parameter to the 3 character team id.

Added statcast_pitcher_team() to statcast_pitcher.py to pull in all pitchers for a specified team during the time range.  Similar to statcast_batter() but changed the player_id parameter to the 3 character team id.

### DIFF
--- a/pybaseball/statcast_batter.py
+++ b/pybaseball/statcast_batter.py
@@ -8,6 +8,29 @@ from . import cache
 from .utils import sanitize_input, split_request
 
 
+def statcast_batter_team(start_dt: Optional[str] = None, end_dt: Optional[str] = None,
+                         team: Optional[str] = None) -> pd.DataFrame:
+    """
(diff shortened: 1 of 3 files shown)


schorrm:
This seems reasonable to me. @TheCleric @tjburch  if you guys agree, then let's add it, subject to adding unit tests for this

tjburch:
Definitely like the idea, but I'm not sure I love introducing an all new function. We could introduce this as a new argument to the current `statcast_batter` function, defaulting to `None` and then make the asserts and scraping behavior based on which is not `None`.  Thoughts? Feel free to override me if you disagree :)

schorrm:
Agreed with Tyler

Issue #279, opened 2022-07-21
Typo in first_name column name for statcast_batter_exitvelo_barrels


miltiades-the-general:
The column titles for the statcast_batter_exitvelo_barrels has a space before the first_name index. It took me a while to realize what was happening when I was trying to reorganize my data. 
![image](https://user-images.githubusercontent.com/97767157/180105408-204284dd-f0f4-4681-8d48-bc10f7020cac.png)

BrayanMnz:
Since this was addressed on https://github.com/jldbc/pybaseball/pull/280/
and merged I think you could close this one also. 

cc: @tjburch @schorrm

tjburch:
Until PyPi picks up 2.2.3 rather than 2.2.1 I'm erring on the side of caution closing issues which are still present in 2.2.1, so people can more easily reference problems they may run into.

tjburch:
Resolved and in pypi following closing of #281.

PR #280, opened 2022-07-27
Sanitize whitespace in column names from statcast calls

tjburch:
This addresses #279, in which there's a whitespace in `first_name`. The issue was flagged for `statcast_batter_exitvelo_barrels`, however it permeated to most the statcast calls. I implemented an effectively one-line solution in utils that strips any whitespace from the column names of the statcast calls, and imported the new function and used where needed.

### DIFF
--- a/pybaseball/statcast_batter.py
+++ b/pybaseball/statcast_batter.py
@@ -5,7 +5,7 @@ import pandas as pd
 import requests
 
 from . import cache
-from .utils import sanitize_input, split_request
+from .utils import sanitize_input, split_request, sanitize_statcast_columns
 
 
 def statcast_batter(start_dt: Optional[str] = None, end_dt: Optional[str] = None, player_id: Optional[int] = None) -> pd.DataFrame:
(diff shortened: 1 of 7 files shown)


tjburch:
Hm. Seems to be failing the test `test_statcast_pitcher_exitvelo_barrels` (and similar tests) on:

`    assert len(result.columns) == 19`

Seems to be getting 18 columns now.

I just tried commenting out my changes to see if I could get back the 19 and still managed to get 18. Can do some more digging into this later but if anyone has an idea where the 19 came from, that'd be helpful. For reference I see:

```
['last_name', 'first_name', 'player_id', 'attempts', 'avg_hit_angle',
       'anglesweetspotpercent', 'max_hit_speed', 'avg_hit_speed', 'fbld', 'gb',
       'max_distance', 'avg_distance', 'avg_hr_distance', 'ev95plus',
       'ev95percent', 'barrels', 'brl_percent', 'brl_pa']
```

tjburch:
Ok, so it seems like even in the last version the Statcast calls only return 18 columns:

```
>>> from pybaseball.statcast_pitcher import statcast_pitcher_exitvelo_barrels
>>> from importlib.metadata import version
>>> version('pybaseball')
'2.2.1'
>>> df = statcast_pitcher_exitvelo_barrels('2020')
>>> len(df.columns)
18
>>> df.columns
Index(['last_name', ' first_name', 'player_id', 'attempts', 'avg_hit_angle',
       'anglesweetspotpercent', 'max_hit_speed', 'avg_hit_speed', 'fbld', 'gb',
       'max_distance', 'avg_distance', 'avg_hr_distance', 'ev95plus',
       'ev95percent', 'barrels', 'brl_percent', 'brl_pa'],
      dtype='object')
```

So I'm just going to update the tests and move on with life

PR #283, opened 2022-08-05
Work in progress: Add functions to fetch FanGraphs projection tables

andersonfrailey:
I started working on a PR to fetch the various [player projections tables from FanGraphs](https://www.fangraphs.com/projections.aspx?pos=all&stats=bat&type=fangraphsdc&team=0&lg=all&players=0) and could use some help. Figured it'd be easier to open a PR than try and explain it all in an issue.

First, I can get all of the data, but not the player names. So the results look like:
Screen Shot 2022-08-05 at 6 28 51 PM
 Would this be fixed by creating a file like `pybaseball/pybaseball/enums/fangraphs/pitching_data_enum.py` that specifies all of the variables? I can't say I fully understand all of the how that works.

Second, I can only get the first 30 players. I see in other FanGraphs functions theres a `page` argument that is added to the URL options passed to the request. However, that doesn't seem to work for the projections tables. If you look in the URL for those pages, there's no parameter for page size. It always
`https://www.fangraphs.com/projections.aspx?pos=all&stats=bat&type=fangraphsdc&team=0&lg=all&players=0`.
I'm nearing the limits of my knowledge about web scraping so I'm not sure if there's another way to handle it.

Sample code for how to use these functions:

```python
from pybaseball import fg_batting_projections_data, fg_pitching_projections_data

pitching_projections = fg_pitching_projections_data('depth_charts')
batting_projections = fg_batting_projections_data('depth_charts')
```
Would fetch the depth chart projections.

Any tips for the two issues I'm running into would be appreciated. Also happy to make changes as requested!


blacktj:
Hi Anderson, 

Are you still working on this and/or want to work on it? I have a solution for it.. I want to sit down and have someone else test it before I propose a crazy change to how one can pull FG data. 
I was able to pull the projections with 1 line of code, but want to make sure we can use it. 

Screen Shot 2023-02-05 at 2 59 17 PM

andersonfrailey:
Thanks for working on it, @blacktj! This is definitely something I still want to work on. I'd be happy to test the solution you came up with

blacktj:
I sent an email to your @uva email.. let me know if you didn't get it. It's a bit wordy for PR chat. I think it checks most of your boxes and it's quite simple.

PR #289, opened 2022-10-15
Issue 284 top prospects

KyleDirks:
I looked up each team's id currently being used on the mlb site and saved as csv to the data folder. Then, I updated teamid_lookup.py with a function to pull these team IDs and referenced that in top_prospects.py to fix the bug


KyleDirks:
Example of using mlb_com_team_id():
```
from pybaseball import teamid_lookup
id = teamid_lookup.mlb_com_team_id(teamName = "bluejays")
print(id)
```

Use of top_prospects is unchanged. Example:
```
from pybaseball import top_prospects
tp = top_prospects("mariners", "pitchers")
tp.head(2)
```
image

tjburch:
Thanks for the contribution @KyleDirks. This already looks pretty good, I'll give it review as soon as I can.

pybaseball/teamid_lookup.py
```diff
@@ -33,6 +33,26 @@ def team_ids(season: Optional[int] = None, league: str = 'ALL') -> pd.DataFrame:
...
 
 
+def mlb_com_team_id(teamName = None):
+    
+    '''
+        Provides team ids use within mlb.com urls
+        
+        if no team is given all team ids are returned in a dataframe; else a single team id is returned as a string
+    '''
+
```
tjburch, line 43: Can you sanitize the input? I'd like to see `bluejays`, `blue_jays`, `blue jays`, or `Blue Jays` all work here. I think this should just be a string lower and remove spaces and underscores.

pybaseball/teamid_lookup.py
```diff
@@ -33,6 +33,26 @@ def team_ids(season: Optional[int] = None, league: str = 'ALL') -> pd.DataFrame:
     return fg_team_data
 
 
+def mlb_com_team_id(teamName = None):
```
tjburch, line 36: Nitpicky: I think `mlb_com_team_id` is going to be hard to remember. Prefer `mlb_team_id` or `mlbam_team_id` personally

tjburch, line 36: Broader point - I think it might be nice to merge this onto the existing `team_ids` function rather than create a new one. Curious what you think (and other devs). 

I think you'd need to add another column for `teamID : team_name`, but then you could merge that onto the dataset produced by that existing function rather than adding a new one.

pybaseball/top_prospects.py
```diff
@@ -3,11 +3,18 @@
 
 from . import cache
 
+from pybaseball import teamid_lookup
+
 
 @cache.df_cache()
 def top_prospects(teamName=None, playerType=None):
-    teamUrl = "" if teamName == None else teamName.lower() + '/'
```
tjburch, line 9: Ah, I see the sanitizing here, which is also nice, but would be cleaner to rope into the function where the teamID is actually gathered so if you want to just import that, you can.

tjburch (commented):
Thanks again for this contribution. Added some comments - if any other devs happen to have a sec to comment it might be useful (@schorrm?)

pybaseball/teamid_lookup.py
BrayanMnz: Agree, changing the name to `mlb_team_id` will give a more intuitive name to the function.

BrayanMnz: I'd rather creating a new function (as it is now in this Pull Request) instead of adding this new to the existing one because in that way we can keep things modularity isolated and remember that the less arguments the function needs is better according to Uncle Bob on `Clean Code`

tjburch:
Closing, covered by #307

PR #291, opened 2022-11-05
Statcast Arm Strength function

wfordh:
Function for pulling data from Statcast's new arm strength leaderboard. Everything ran with how I tested locally. I kept with how I did the other functions and left out any ability to switch to the team rollups that are possible on the actual site. There's also a weird thing where the infield (2B/SS/3B) group is labeled as "inf" in the URL instead of "if" for some reason, but that seems like another Statcast oddity. Happy to handle it another way.

### DIFF
--- a/pybaseball/statcast_fielding.py
+++ b/pybaseball/statcast_fielding.py
@@ -55,3 +55,18 @@ def statcast_catcher_framing(year: int, min_called_p: Union[int, str] = "q") -> 
 	# CSV includes league average player, which we drop from the result
 	return data.loc[data.last_name.notna()].reset_index(drop=True)	
 
+@cache.df_cache()
+def statcast_arm_strength(year: int, pos: Union[int, str], min_throws: int = 100):
+	# want the pos as a "code" not a name or number
+	pos = norm_positions(pos, to_word=False, to_number=False)
+	# The 2B/SS/3B option uses "inf" instead of "if" as an abbreviation for
+	# "infield" for some reason
+	if pos == "if":
+		pos = "inf"
+	if pos == "2":
+		raise ValueError("This particular leaderboard does not include catchers!")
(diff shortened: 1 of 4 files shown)


pybaseball/statcast_fielding.py
```diff
@@ -55,3 +55,18 @@ def statcast_catcher_framing(year: int, min_called_p: Union[int, str] = "q") ->
 	# CSV includes league average player, which we drop from the result
 	return data.loc[data.last_name.notna()].reset_index(drop=True)	
 
+@cache.df_cache()
+def statcast_arm_strength(year: int, pos: Union[int, str], min_throws: int = 100):
```
bdilday, line 59: I think you've got `pos` and `min_throws` in a different order than in the readme or the tests. 
`

wfordh: That seems to have been it, but still failing the tests. I searched the logs and didn't find any mention of failing this particular test. Am I missing other changes or something?

pybaseball/statcast_fielding.py
```diff
@@ -55,3 +55,18 @@ def statcast_catcher_framing(year: int, min_called_p: Union[int, str] = "q") ->
...
+	# want the pos as a "code" not a name or number
+	pos = norm_positions(pos, to_word=False, to_number=False)
+	# The 2B/SS/3B option uses "inf" instead of "if" as an abbreviation for
+	# "infield" for some reason
+	if pos == "if":
+		pos = "inf"
+	if pos == "2":
+		raise ValueError("This particular leaderboard does not include catchers!")
+	url = f"https://baseballsavant.mlb.com/leaderboard/arm-strength?type=&year={year}&minThrows={min_throws}&pos=arm_{pos}&team=&csv=true"
```
wfordh, line 68: Right now since `to_number=False` for `norm_positions`, the URL includes `pos=arm_all` for the `pos` parameter. On the actual leaderboard `pos=` is how it works for getting all positions and the current solution redirects to that URL, so it returns the same thing in the end. Not sure if I should handle that a different way, can try tinkering with `norm_positions()` if we want.

Issue #281, opened 2022-07-27
PyPi still using release 2.2.1


tjburch:
If you want the fresh 2.2.3, you'll have to install from the repo. 

Brought up in the devs slack but just wanted to make an issue about it so we could remember/track it, and so it'd be out there in the public too. Not sure what needs to be done workflow-wise here but just keeping it on the radar.

bdilday:
I juts noticed that the deploy stage of the CI failed https://github.com/jldbc/pybaseball/actions/runs/2584633111
The logs have expired so not sure why. Maybe we could bump to `2.2.4` and try to release to find out what the error was?

tjburch:
Good catch - @schorrm are you able to do that?

schorrm:
Yes. Will do today

tjburch:
I think we just need to bump the number here: https://github.com/jldbc/pybaseball/blob/9c0939020659b1d015f429031fce7a74923e9f16/setup.py#L22

tjburch:
[Updated](https://github.com/jldbc/pybaseball/commit/983902b7821659d34fe8755822298ceb335b3051). Tried rerunning the most recent 2.2.4 release workflow but still got the failure. I _think_ it's just re-running the action as-is, without picking up the newest file change. Which means we hopefully just need to manually set off another call to that workflow. Worst case scenario, change setup.py to 2.2.5 and have yet another release.

PR #295, opened 2022-12-10
Add module level __version__method

tjburch:
Following some of the annoyance in #281, I thought it might be good to introduce a module level __version__ method like a lot of packages out there have. This adds that to a new file called `pybaseball/version.py`, which we should increment at every release. It's loaded in `pybaseball.__init__.py` so you can do the following:

```
>>> import pybaseball
>>> pybaseball.__version__
'2.2.4'
```

Then lastly, `setup.py` loads in the value from that file rather than internally defining it.


BrayanMnz:
Safe to merge per my review.

tjburch:
With all the recent bug fixes, I think it's time to version bump. I raised the value on this PR, and then we should be good to go for that. @schorrm, if you're good with this, merge and bump. Otherwise if you don't want the method in this PR you can just raise the numbers without the new method and bump.

tjburch:
replaced by #320 after a truly wretched diff here.

PR #296, opened 2022-12-11
Add 3.10 and 3.11 and remove 3.6 and 3.7 from test versions

schorrm:
* Switching to quotes, as is now necessary for 10/11 and might as well be uniform.
* Removing 3.6 and 3.7 as our dependencies have dropped support and we can't feasibly offer it.
* Go nuts on the walrus operator now I guess?


tjburch:
👍

Should also edit setup.py correspondingly to drop 3.6 and 3.7.

TheCleric (approved):

tjburch:
Is this good to go? @schorrm

PR #292, opened 2022-12-02
Add discord link (and new version number)

tjburch:
Adds new discord link to README. Won't merge until a couple developers join in to help with admin stuff. Also raises the version number on the readme to 2.3.3.

Will close #277.

PR #297, opened 2022-12-11
README rewrite

tjburch:
Overhaul README to be somewhat less verbose and to-the-point. Aimed to reduce the number of full dataframe printouts we had, raise the full docs folder to be more visible. Some notes:

- Removed the version number. It's on git, so it should be clear by the sidebar
- Removed the changelog link. Git has an internal changelog that we seem to be better about using.
- I think it makes sense for contributing.md to be a contributing guide rather than suggestions on things to add. So I wrote it as if that's the future paradigm. We'll need to update contributing.md correspondingly.

Has the discord link so supersedes #292. I'll take that one down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ pip install -e .
 
 We will try to publish periodic updates through the 'releases' and PyPI CI, but it may lag at times.
 
+## Community
+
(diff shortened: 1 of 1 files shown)


schorrm:
LGTM in general, I would put the discord link at the top though

PR #299, opened 2022-12-12
Feat: Update "OOA Behind Pitcher" function

zach-hopkins:
Proposed feature which simply adds an argument to the "pitcher_fielding" statcast pull in order to allow for "view" changes (options are: Fielder, Fielding Team, Batting, Batting Team, Pitcher). Helpful for pulling OOA and RAA for defense behind a particular pitcher and retrieving team fielding information. Set default to previous hardcode "Fielder"

### DIFF
--- a/pybaseball/statcast_fielding.py
+++ b/pybaseball/statcast_fielding.py
@@ -8,12 +8,12 @@ from . import cache
 from .utils import norm_positions, sanitize_statcast_columns
 
 @cache.df_cache()
-def statcast_outs_above_average(year: int, pos: Union[int, str], min_att: Union[int, str] = "q") -> pd.DataFrame:
+def statcast_outs_above_average(year: int, pos: Union[int, str], view: str = "Fielder", min_att: Union[int, str] = "q") -> pd.DataFrame:
 	pos = norm_positions(pos)
 	# catcher is not included in this leaderboard
 	if pos == "2":
 		raise ValueError("This particular leaderboard does not include catchers!")
-	url = f"https://baseballsavant.mlb.com/leaderboard/outs_above_average?type=Fielder&year={year}&team=&range=year&min={min_att}&pos={pos}&roles=&viz=show&csv=true"
+	url = f"https://baseballsavant.mlb.com/leaderboard/outs_above_average?type={view}&year={year}&team=&range=year&min={min_att}&pos={pos}&roles=&viz=show&csv=true"
 	res = requests.get(url, timeout=None).content
 	data = pd.read_csv(io.StringIO(res.decode('utf-8')))
 	data = sanitize_statcast_columns(data)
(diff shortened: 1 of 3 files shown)


schorrm:
Looks good.
1. Move the new argument to be last for now? @tjburch do we care that much about breaking compat here
2. Add to tests?
3. Is there anything that needs updates for the docs?

pybaseball/statcast_fielding.py
```diff
@@ -8,12 +8,12 @@
 from .utils import norm_positions, sanitize_statcast_columns
 
 @cache.df_cache()
-def statcast_outs_above_average(year: int, pos: Union[int, str], min_att: Union[int, str] = "q") -> pd.DataFrame:
+def statcast_outs_above_average(year: int, pos: Union[int, str], view: str = "Fielder", min_att: Union[int, str] = "q") -> pd.DataFrame:
 	pos = norm_positions(pos)
```
tjburch, line 12: We should have a docstring here in the first place, but it would be nice to include the options for `view` in that docstring. If you could start with the following and add the options, that would be awesome.

```
	"""Scrapes outs above average from baseball savant for a given year and position

	Args:
		year (int): Season to pull
		pos (Union[int, str]): Numerical position (e.g. 3 for 1B, 4 for 2B). Catchers not supported
		min_att (Union[int, str], optional): Integer number of attempts required or "q" for qualified. 
			Defaults to "q".

	Raises:
		ValueError: Failure if catcher is passed

	Returns:
		pd.DataFrame: Dataframe of defensive OAA for the given year and position for players who have met
			the given threshold
	"""
```

tjburch:
Hey, thanks Zach, this is a clever addition to make this more useful.

Agree with @schorrm on moving the argument to the end. I added a review about a docstring so we know what all can be put there.

For the test, make sure [this one](https://github.com/jldbc/pybaseball/blob/5869a2d9c5cd1f8defc3b5da91d5d5ca5dc5ebb5/tests/integration/pybaseball/test_statcast_fielding.py#L12) still runs with the default value, then add a second just like it for a different option of `view`.

For the docs, just add a note in [this](https://github.com/jldbc/pybaseball/blob/master/docs/statcast_fielding.md) file with the alternate options for views.

Thanks again!

zach-hopkins:
> Hey, thanks Zach, this is a clever addition to make this more useful.
> 
> Agree with @schorrm on moving the argument to the end. I added a review about a docstring so we know what all can be put there.
> 
> For the test, make sure [this one](https://github.com/jldbc/pybaseball/blob/5869a2d9c5cd1f8defc3b5da91d5d5ca5dc5ebb5/tests/integration/pybaseball/test_statcast_fielding.py#L12) still runs with the default value, then add a second just like it for a different option of `view`.
> 
> For the docs, just add a note in [this](https://github.com/jldbc/pybaseball/blob/master/docs/statcast_fielding.md) file with the alternate options for views.
> 
> Thanks again!

No problem! Added the changes, let me know if all looks right. Works well for on my end.

schorrm:
LGTM for me

tjburch:
And also me. Seems pointless to run tests right now since they're all broke anyway so smashing that merge button.

Thanks a ton @zach-hopkins!

PR #307, opened 2023-01-21
Issue 284 top prospects not working.

bryanpeabody:
Finished the work KyleDirks started and added sanitization around team names, blue Jays, blue-jays, etc are changed to lowercase bluejays so all combinations work. Added some error checking for empty team names as well.


BrayanMnz:
It seems ok for me, 
Could be merged per my revision. 
Thoughts ? @tjburch, @schorrm

pybaseball/teamid_lookup.py
```diff
@@ -33,6 +33,26 @@ def team_ids(season: Optional[int] = None, league: str = 'ALL') -> pd.DataFrame:
...
+    mlb_url_team_ID = pd.read_csv(MLB_URL_FILE_NAME, index_col=0)
+    
+    if teamName is not None and len(teamName.strip()) != 0:
+        # Sanitize the input - remove spaces, dashes and make lower case
+        teamName = teamName.replace(" ", "").replace("-", "").lower()
+
+        mlb_url_team_ID = mlb_url_team_ID.query(f"team_name == '{teamName}'")
+        mlb_url_team_ID = mlb_url_team_ID['mlb_team_id'].loc[mlb_url_team_ID.index[0]]
+
+    return mlb_url_team_ID
```
tjburch, line 53: This is kind of nit-picky but I'm not sure I _love_ that this method returns either a string or a dataframe. The methods and approaches you would do to a string are pretty different from that you would do on a DF. IMO I'd add a second function that just returns just the DF and then let `mlb_team_id` slice the DF to the passed teamName

tjburch (commented):
This looks really good, thanks for taking care of it. I left only one piece of feedback (that I don't feel overly strong about).

pybaseball/teamid_lookup.py
bryanpeabody: @tjburch I initially just used the PR/code changes that someone else had submitted so we could get the fix out. I agree that function needed some work. So, I went ahead and cleaned it up. It will now either return a number such as 112 (if the string Cubs is passed it) or throw an error if the team name isn't found.

tjburch (commented):
LGTM. @schorrm if you want to give this a look over and merge it when you have a chance. Could be done before version bumping so we get it in.

PR #311, opened 2023-01-31
Dropping Python 3.6 from python-versions

BrayanMnz:
Since we have dropped Python 3.6 support on the PyBaseball project - I'm dropping the 3.6 version of Python from the python-version on the test jobs strategy.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,9 +12,10 @@ jobs:
     strategy:
       matrix:
         python-version:
-        - 3.7
         - 3.8
         - 3.9
+        - 3.10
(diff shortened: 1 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -12,7 +12,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-        - 3.6
```
schorrm, line 15: remove 3.7 and add:

        - "3.8"
        - "3.9"
        - "3.10"
        - "3.11"

schorrm (changes requested):

PR #312, opened 2023-01-31
Create issue templates for PyBaseball project.

BrayanMnz:
Creating the template for PyBaseball project, 
initially we're just using some of the default options provided by Github.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
```
tjburch, line 23: I don't know about others but I personally dislike screenshots in issues. Would rather see the copy/pasted error message as-is.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
```
tjburch, line 20: IMO I think the first three forms here are going to most often yield the same thing 3 times over and over: "I expected this to happen when I did this." I'd prefer one simpler form, something like 

```
***Describe the bug and how to produce it***
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: PYBASEBALL-BUG-
+labels: ''
+assignees: ''
```
tjburch, line 6: I'd remove assignees. If things get assigned, I think that should be done by maintainers or whomever wants to do it, I'd rather not have the submitter choose who.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. Ubuntu, Windows]
+ - Python version [e.g. 3.7, 3.9]
+
```
tjburch, line 29: Add:

```
- pybaseball version (e.g. 2.2.3, 2.2.4, or development if a direct clone)
```

I think this also motivates merging #295 to make this more accessible.

tjburch (commented):
My personal ethos is that we should make it as easy as possible for people to submit an issue, and these have a tendency to add another hoop to jump through, so I'm not 100% sold on issue templates. However, if the community wants it, I'm happy to add it.

I think removing some of the redundancy as mentioned in the review so it's just very quick "tell me your problem and your setup" would help it feel less like a hurdle.

PR #316, opened 2023-02-07
Updating top_prospects to allow for FanGraphs datasource

blacktj:
Updating top_prospects to allow for FanGraphs datasource 
- Add docstring 
- Params added: fg, year with defaults 
- Add fg if/else block 
- Deprecated teamUrl in MLB scrape block

I wanted to place a small commit in a "low-use" file show the use of the FG api. I've started using it instead of some of the tools embedded here and have found it to be solid.

### DIFF
--- a/pybaseball/top_prospects.py
+++ b/pybaseball/top_prospects.py
@@ -1,28 +1,65 @@
 import pandas as pd
 import requests
+import json
+from datetime import date
 
 from . import cache
 
 
 @cache.df_cache()
-def top_prospects(teamName=None, playerType=None):
-    teamUrl = "" if teamName == None else teamName.lower() + '/'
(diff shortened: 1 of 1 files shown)


blacktj (commented):
Sorry for this one. 

I removed the postprocess function in my commit: 



```
def postprocess(prospectList):
    prospectList = prospectList.drop(list(prospectList.filter(regex = 'Tm|Unnamed:*')), axis = 1)
    return prospectList
```

blacktj:
This is going to run into a conflict with the other top prospects PRs that have been discussed.. I may just separate it into a different function to keep it clean. Thoughts?

blacktj:
I'll rebuild this to allow for the mlb datasource to be refactored as well.

PR #334, opened 2023-03-08
Add Negro Leagues To `FangraphsLeague`

nwinston:
Now that [Fangraphs supports negro leagues](https://blogs.fangraphs.com/negro-leagues-data-is-now-available-on-fangraphs/), we should be able to pull that data through pybaseball. New values need to be added to the `FangraphsLeague` enum otherwise an exception is thrown.

### DIFF
--- a/docs/batting_stats.md
+++ b/docs/batting_stats.md
@@ -9,7 +9,7 @@ The `batting_stats` function returns season-level batting data from FanGraphs.
 
 `end_season:` Integer. The final season you want to retrieve data from. If omitted, the funciton will return only data from the start_season.
 
-`league:` String. Either "all" for all data, "nl" for National League, "al" for the American League, or "mnl" for all Negro League data. Defaults to "all", for returning data on all teams.
+`league:` String. Either "all" for all data, "nl" for National League, "al" for the American League, or "mnl" for all Negro League data. Defaults to "all", for returning data on all teams. See `FangraphsLeague` definition for all leagues.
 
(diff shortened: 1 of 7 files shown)


BrayanMnz:
Looks good to me. @tjburch

tjburch:
thanks @nwinston, this will be really good to add. 

Do you think you could add comments to the docs where this might be used? E.g. in [team_batting.md](https://github.com/jldbc/pybaseball/blob/0cdbe5d8da89211642dddee9610514c43c5166e1/docs/team_batting.md?plain=1#L12) could change 

``league:` String. Either "all", "nl", or "al" for determining whether you want data from one league or both. Defaults to "all", for returning data on all teams. `

to 

``league:` String. Either "all" for all data, "nl" for the National League, "al" for the American League, or "mnl" for the Negro Leagues. Defaults to "all", for returning data on all teams. `

nwinston:
Done! @tjburch

tjburch:
Thanks. LGTM! Thanks a ton. Will give a day for anyone else to make comments and merge after. ( @schorrm )

schorrm:
Looks amazing! (thanks for tagging me @tjburch)
One question though: is there a reason why the param can only be 'mnl'?

nwinston:
Updated the docs in `fangraphs.md` that says to look at `FangraphsLeague` for the full set. I'll include that note in the rest of them as well.

schorrm:
Anyway, more than LGTM, @tjburch merge whenever you want

tjburch:
In. Thanks again @nwinston 👍👍

nwinston:
No problem!

PR #335, opened 2023-03-10
Scrape Player Projection Data from Fangraphs

TK2575:
Introduces a function and related tests and documentation that captures player projection data from Fangraphs. Provides argument options to specify the projection source, position, league and team. Extends the teamid lookup method to provide a fg team ID lookup needed for applying team level filtering using a stored dictionary fixture.


docs/fangraphs.md
```diff
@@ -1,4 +1,4 @@
-# Fangraphs Data Acquisition Functions
+# Fangraphs Player and Team Performance Data Acquisition Functions
```
TK2575, line 1: Renamed to clarify difference between performance status and projection status

pybaseball/fangraphs_projections.py
```diff
@@ -0,0 +1,112 @@
...
+    Returns:
+        pd.DataFrame: Projections data set given applied filters, columns vary for batters versus pitcher. Returns empty dataframe when no data found on Fangraphs
+    """
+    args = _FgProjectionArgument(position, proj_source, league, team)
+    source = requests.get(args.url, timeout=None)
+
+    soup = BeautifulSoup(source.content, 'html.parser')
+    raw = soup.find('script', id='__NEXT_DATA__', type='application/json').string
+    data = json.loads(raw)['props']['pageProps']['dehydratedState']['queries'][0]['state']['data']
+    result = pd.read_json(json.dumps(data))
```
TK2575, line 30: Turns out the entire projection payload is present in the page source even if the default view only displays the first 30 rows

blacktj: I think you're doing too much work here there's an api: 
pitch_df = pd.DataFrame(json.loads(requests.get('[https://www.fangraphs.com/api/projections?stats=pit&type=steamer').content]

blacktj:
Sorry commented in line as opposed to on PR: 

I think you're doing too much work here there's an api:
`pitch_df = pd.DataFrame(json.loads(requests.get('[https://www.fangraphs.com/api/projections?stats=pit&type=steamer').content]`

TK2575:
Thanks @blacktj, didn't know that API endpoint existed in front of a paywall, that's great! I'm assuming there's some rate limit expectation we'll need to respect like we do with baseball reference? I'll need to dig into this a bit.

blacktj:
It's non-public and buried in the client-side rendering of the table. I am working on a PR for the prospects endpoint of this as well.. not sure if it's rate limited though. It's wide open. The risk I see is if they do lock it down.

TK2575:
I think I'll need to defer to this repo's maintainers as to which approach to take. There's precedence for scraping Fangraphs page source for other methods, though I don't know if that's because either a) we weren't aware of the API at the time or b) the API didn't/doesn't support those data. Querying from the API would certainly be cleaner, but I'd be hesitant in moving forward using a non-public API without some form of developer contract and/or buy-in from this repo's maintainers.

blacktj:
This is a webscraping repo.. so I'm guessing we don't have a contract to pull the data from their actual website? Is there a difference between grabbing it there or from the API they use to render the table?

pybaseball/fangraphs_projections.py
TK2575: Per @blacktj we may want to utilize non-public but otherwise accessible API to capture this data directly: 
```
json.loads(requests.get('https://www.fangraphs.com/api/projections?stats=pit&type=steamer').content
```

JamesYouL2: > 

You just saved me $10/month and made my life a ton easier.

PR #366, opened 2023-06-09
Update fangraphs month enum to support 'last x' games

abrobert:
Fangraphs allows filtering leaderboards by last seven, fourteen and thirty days by setting the month to 1, 2, or 3 respectively.

- Adds support to pybaseball to filter by these values from fangraphs


abroberts14:
Any update on this? @markspotsthex

Issue #375, opened 2023-08-16
statcast throwing KeyError on certain dates in 2023


Haman-Karn:
While getting all of the statcast data, I kept getting an error around 98%. So I eventually was able to narrow it down to 2023-06-25 being the first problematic one day. Other day(s) past this one also cause the error, but I've stopped at 06-25 because this amount of data is good enough for my current purposes.

The code I'm executing is this:
`stats = statcast(start_dt="2023-06-25")`

Upon execution, my terminal looks like this:
```
This is a large query, it may take a moment to complete
  0%|                                                                                                                                   | 0/1 [00:00<?, ?it/s] 
Traceback (most recent call last):
  File "c:\Users\nosoa\Documents\glb\getstats.py", line 6, in <module>
    stats = statcast(start_dt="2023-06-25")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\nosoa\Documents\glb\venv\Lib\site-packages\pybaseball\statcast.py", line 113, in statcast
    return _handle_request(start_dt_date, end_dt_date, 1, verbose=verbose,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\nosoa\Documents\glb\venv\Lib\site-packages\pybaseball\statcast.py", line 76, in _handle_request
    dataframe_list.append(future.result())
                          ^^^^^^^^^^^^^^^
  File "C:\Users\nosoa\AppData\Local\Programs\Python\Python311\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\nosoa\AppData\Local\Programs\Python\Python311\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Users\nosoa\AppData\Local\Programs\Python\Python311\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\nosoa\Documents\glb\venv\Lib\site-packages\pybaseball\cache\cache.py", line 58, in _cached
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #379, opened 2023-08-20
Statcast cache incomplete issue

ss77995ss:
### Related Issue
* #375 

### Description
* Cache with `statcast` function will have incorrect value and cause `cache.purge()` not work.
* Also will cause the cache has no expire time and occupied in memory.

### What cause this issue
* If `df_cache` is wrapped by another `df_cache` the parent cache will finish writing once the child cache start writing.
* Parent cache file will become invalid JSON file and cannot be parsed correctly.

### How to fix
* Remove the `df_cache` for `_small_request`


ss77995ss:
Fixed in #438

Issue #387, opened 2023-10-04
Refactor Print statements to use python logging module


pinsondg:
Whenever I go to fetch data from a statcast function, there is a print statement that is being run ([example](https://github.com/jldbc/pybaseball/blob/e94093b3985f4bbfa88b3c413f16391686163abd/pybaseball/utils.py#L284C1-L284C1)). This is clogging up my output and it would be better if the [python logging module](https://docs.python.org/3/library/logging.html) was used instead so it can be easily disabled. There are print statements in a few places that should also be changed. This will allow the introduction of log levels as well.

PR #388, opened 2023-10-04
Logging improvements

pinsondg:
Replacing print statements with logger module


samlafell:
Hey @pinsondg, was there an issue for this? Ideally instead of directly submitting a PR, we would submit an issue, get some eyes on it, and then determine if a PR is needed and if someone should be working on it.

If not, I guess I'll let the maintainers determine what their thoughts are for incorporating this later in a PyPI update.

pinsondg:
Here is the [issue](https://github.com/jldbc/pybaseball/issues/387)

Issue #393, opened 2023-11-30
Player names in Baseball Reference stats contain mis-encoded non-ASCII characters


AndrewsOR:
The FanGraphs functions `pitching_stats()` and `batting_stats()` appear to convert names from that [site](https://www.fangraphs.com/leaders/major-league?pos=all&stats=bat&lg=all&qual=y&type=8&season=2023&month=0&season1=2023&ind=0) such as Ronald Acuña Jr. and José Abreu to `Ronald Acuna Jr.`, `Jose Abreu` etc., which are recognizable if not entirely correct.

On the other hand, the Baseball Reference functions `batting_stats_bref()` and `pitching_stats_bref()` return what seems like mis-converted HTML encodings of those names, resulting in lower readability, although the names on the [site](https://www.baseball-reference.com/players/a/acunaro01.shtml) itself appear correct.

For example:

```
import pybaseball as pb
df_bref_batting = pb.batting_stats_bref(2023).sort_values("mlbID")
df_bref_pitching = pb.pitching_stats_bref(2023).sort_values("mlbID")

for side, df in zip(["batting","pitching"],[df_bref_batting,df_bref_pitching]):
    print(side)
    print(df[df["Name"].str.contains("x")][["Name","mlbID"]].head().to_string(index=False)+"\n")
```
prints:

```
batting
                        Name  mlbID
           Manny Pi\xc3\xb1a 444489
     Mart\xc3\xadn Maldonado 455117
           Sandy Le\xc3\xb3n 506702
Avisa\xc3\xadl Garc\xc3\xada 541645
         Carlos P\xc3\xa9rez 542208

pitching
                   Name  mlbID
           Max Scherzer 453286
Mart\xc3\xadn Maldonado 455117
     Luis Garc\xc3\xada 472610
   Jos\xc3\xa9 Quintana 500779
              Alex Cobb 502171
```

My current workaround is to use `playerid_reverse_lookup` to bridge to FanGraphs names and use those instead. (I like to use the Baseball Reference batting stats because of how it labels players who played in multiple teams/leagues in a given season, providing both team names instead of "---".)

I love `pybaseball`... thank you!

BrayanMnz:
Hi, these are "tildes" for spanish words, 
I have a workaround for this in an internal project that maybe can be useful. 

Will try to replicate your example and apply my workaround

BrayanMnz:
The issue is because we are wrongly encoding a _**bytes**_ object parsed to _**string**_ 
[...]

PR #394, opened 2023-12-01
ISSUE 393 : Decoding the content into utf-8 to avoid encoding issues 

BrayanMnz:
The issue is that we are wrongly encoding a bytes object parsed to a string
what we need to do instead is, decode the bytes object directly. 

This commit changes that, to decoding the content directly from _**bytes**_ into utf-8

See, the results after these changes: 

https://github.com/jldbc/pybaseball/issues/393#issuecomment-1836852467

### DIFF
--- a/pybaseball/league_batting_stats.py
+++ b/pybaseball/league_batting_stats.py
@@ -20,7 +20,7 @@ def get_soup(start_dt: date, end_dt: date) -> BeautifulSoup:
     url = "http://www.baseball-reference.com/leagues/daily.cgi?user_team=&bust_cache=&type=b&lastndays=7&dates=fromandto&fromandto={}.{}&level=mlb&franch=&stat=&stat_value=0".format(start_dt, end_dt)
     s = session.get(url).content
     # a workaround to avoid beautiful soup applying the wrong encoding
-    s = str(s).encode()
+    s = s.decode('utf-8')
     return BeautifulSoup(s, features="lxml")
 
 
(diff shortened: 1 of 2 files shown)


bdilday:
I think this change was on purpose in order to workaround a different issue - see 
https://github.com/jldbc/pybaseball/pull/223
https://github.com/jldbc/pybaseball/issues/218

you may want to double check that this PR doesn't reintroduce the intermittent truncated dataframe issue mentioned in https://github.com/jldbc/pybaseball/issues/218

BrayanMnz:
Could you elaborate on how to replicate the old issue? @bdilday 

Just to validate this not introduce a regression

BrayanMnz:
Hi, @bdilday 

Replicating the old issue, there is no data truncated with my fix. 

I used the same example as the previous issue 

https://github.com/jldbc/pybaseball/assets/61812255/97706506-05dc-4bf5-92c6-3a4a7a799076

bdilday:
> Could you elaborate on how to replicate the old issue? @bdilday
> 
> Just to validate this not introduce a regression

it's tricky to replicate because it depends not only on the data but on the (randomly generated) links in the page header. That is, if the page contained a link to the german language or spanish language fbref pages, AND a player in the table had a non-ascii character in their name, then the table got truncated.  

but the problem seems to be resolved now. Here's an example that should reproduce the original issue but works fine 
https://gist.github.com/bdilday/6e6bef77a5e42776a7e6da7bee12178d

bdilday:
> So, my PR should also solve the issue reported on ISSUE 393. Could you review it, again?
> 
> Also tested the snippet you've shared with my changes and is working as expected.
> 
> FYI @bdilday

Yes, I think this PR looks good to go!

Issue #414, opened 2024-04-03
retrosheety.py returns an error for all functions


grantdanielson:
The Chadwick Baseball Bureau seems to have reorganized their enhanced retrosheet repo. This means some modifications need to be made to retrosheet.py.

Here are the errors for each function:
```
events('2023')
  File "...\pybaseball\retrosheet.py", line 130, in events
    subtree = repo.get_git_tree(subtree.sha)
                                ^^^^^^^
UnboundLocalError: cannot access local variable 'subtree' where it is not associated with a value
```
```
rosters('2023')
  File "...\pybaseball\retrosheet.py", line 164, in rosters
    rosters = [t.path for t in repo.get_git_tree(subtree.sha).tree if str(season) in t.path]
                                                 ^^^^^^^
UnboundLocalError: cannot access local variable 'subtree' where it is not associated with a value
```
```
park_codes()
  File "...\pybaseball\retrosheet.py", line 213, in park_codes
    data.columns = parkcode_columns
    ^^^^^^^^^^^^
TypeError: park_codes() takes 0 positional arguments but 1 was given
```
```
schedules('2023'
  File "...\pybaseball\retrosheet.py", line 224, in schedules
    schedules = [f.path[f.path.rfind('/')+1:] for f in repo.get_contents('schedule')]
                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
```
season_game_logs('2023')
  File "...\pybaseball\retrosheet.py", line 246, in season_game_logs
    raise ValueError(f'Season game logs not available for {season}')
ValueError: Season game logs not available for 2023
```
`world_series_logs()`, `all_star_game_logs()`, `wild_card_logs()`, `division_series_logs()`, `division_series_logs()`, and `lcs_logs()` all do not have any output in the terminal and no files are created.

grantdanielson:
All retrosheet functions should work now

Issue #407, opened 2024-03-16
season_game_logs() from retrosheet.py returns error for all seasons


jackspagna:
Using the season_game_logs() function raises error: "ValueError(f'Season game logs not available for {season}')" for all seasons because the season game log files have moved from the gamelog/ folder to the respective season's seasons/{season} folder (seasons/1978, seasons/2023, etc) in the Chadwick Bureau GitHub.

BrayanMnz:
Looks like you know the reason of the issue, why not submitting a Pull request?

PR #415, opened 2024-04-03
Updating retrosheet.py to work with new chadwickbureau/retrosheet repo

### LINKED ISSUE #392: Can't load 2023 retrosheet data
I tried loading retrosheet events for 2023, and got the *ValueError: Event files not available for 2023*.  I see that the 2023 data hasn't been released yet even though the season is over.  Is there any other way to join the statcast data with data about weather, wind, and umpire?  That's primarily what I'm using the data for.

grantdanielson:
Gradually updating retrosheet.py to work with the new organization of Chadwick Bureau's retrosheet repo. As described in the issue (#414), all functions return errors or just simply do not work. This should also eventually resolve #392 & #407.

Initial commits fix the events and season_game_logs functions.


schorrm:
Thank you!

Issue #437, opened 2024-07-04
Cache Generates Massive Amounts of Empty Files


BOBONA:
Using the cache is generating these files. 

_small_request1720064495.401325.cache_record.json

The files contain `{"func": "_small_request", "args": [`

bdilday:
Thanks for raising this! I have been wondering why `cache.purge()` fails to work for me and I think this is the cause. I think the root cause is the cache is trying and failing to serialize dates. I think it is fixed by https://github.com/jldbc/pybaseball/pull/438

PR #438, opened 2024-07-04
cache serializing error

bdilday:
The cache writer has been writing corrupted files. The reason seems to be that statcast takes dates as args and these can't be json serialized, example,

```
>>> import json
>>> import datetime
>>> json.dumps(datetime.date.today())
  File "/usr/lib/python3.12/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type date is not JSON serializable
```

closes https://github.com/jldbc/pybaseball/issues/437

### DIFF
--- a/pybaseball/cache/cache.py
+++ b/pybaseball/cache/cache.py
@@ -1,4 +1,5 @@
 import abc
+import datetime
 import functools
 import glob
 import os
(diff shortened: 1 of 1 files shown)


ss77995ss:
Hi @bdilday . It looks like this PR could also fix this issue! Could you also test this case, too?
If so I am going to close my PR
ref: https://github.com/jldbc/pybaseball/pull/379

PR #435, opened 2024-06-20
Alternative Lahman DB route

jmaslek:
Based on this comment: https://github.com/jldbc/pybaseball/pull/434#issuecomment-2179544540

I just threw the files onto my own repo (can be easily adjusted to be on pybaseball, I jwould just rather it be added so we can point the link at main instead of this branch).

There is one manually maintained file (teams_upstream) that I do not know anything about.

PR #434, opened 2024-06-18
Fix the Lahman Database Scraping

jmaslek:
This PR redirects the Lahman database from a 404 github link to the dropbox site that is found on baseball1.com.

In order to extract the data, py7zr was added to the requirements.

PR #449, opened 2024-09-02
fixed lahman.py; added test_lahman.py

mlinenweber:
- Fixed `lahman.py` - now downloads from dropbox [does so in a generic way, by finding the link in http://lahman.com]
- Added `test_lahman.py`
- Refactored `conftest.py`, by combining `response_get_monkeypatch` and `bref_get_monkeypatch` to use `target_get_monkeypatch` [they differed by only 1 line].
- Added fixtures `get_data_file_bytes` and `target`.
- Updated documentation.
- Updated `setup.py` to install `py7zr` and `requests_cache`


bdilday:
there is also a fix here https://github.com/jldbc/pybaseball/pull/434
I think there's a preference to avoid relying on 7-zip and `py7zr`?

bdilday:
I meant this PR https://github.com/jldbc/pybaseball/pull/435

mlinenweber:
Respectfully, I don't know of any reason why not to use 7zip/py7z. Now, the justification to use it is that it is the compression format Sean Lahman has chosen. Furthermore, to that extent, I don't think it is appropriate to rely upon a third-party to host a zip file. (Same goes for putting the zip file into this repository). I would rather go directly to the source, i.e., [seanlahman.com](http://seanlahman.com/) - that way when he updates the data this library can download the latest without delay.

pybaseball/lahman.py
```diff
@@ -1,136 +1,161 @@
...
+    response = session.get("http://seanlahman.com", refresh=force)
+    return response
+
+# For example, "https://www.dropbox.com/scl/fi/hy0sxw6gaai7ghemrshi8/lahman_1871-2023_csv.7z?rlkey=edw1u63zzxg48gvpcmr3qpnhz&dl=1"
+def _get_download_url(force:bool=False) -> str:
+    response = _get_response(force)
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    anchor = soup.find("a", string="Comma-delimited version")
+    url = anchor["href"].replace("dl=0", "dl=1")
```
SeanCooke, line 27: @mlinenweber Can you explain these 2 lines and what _get_download_url() is supposed to return?

```python
    anchor = soup.find("a", string="Comma-delimited version")
    url = anchor["href"].replace("dl=0", "dl=1")
```

I get the below error when trying to run `people()`

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[24], line 31
     28 soup = BeautifulSoup(response.content, "html.parser")
     30 anchor = soup.find("a", string="Comma-delimited version")
---> 31 url = anchor["href"].replace("dl=0", "dl=1")

TypeError: 'NoneType' object is not subscriptable
```

Is `soup.find("a", string="Comma-delimited version")` supposed to be None?

Issue #467, opened 2025-01-28
repeated warning when running pybaseball.statcast()


johnnynienstedt:
The following warning is raised once for each day in the query: 

FutureWarning: errors='ignore' is deprecated and will raise in a future version. Use to_datetime without passing `errors` and catch exceptions explicitly instead
  data_copy[column] = data_copy[column].apply(pd.to_datetime, errors='ignore', format=date_format)

mattdennewitz:
The warning is worthwhile to keep, and can be suppressed in your implementation like so:

```python
import warnings

with warnings.catch_warnings():
    warnings.simplefilter("ignore", FutureWarning)

    df = statcast(...)
```

PR #474, opened 2025-03-31
Fix datetime deprecation

hsum:
## Fix pandas to_datetime deprecation warning

This PR removes a FutureWarning in pybaseball/datahelpers/postprocessing.py related to the use of `errors='ignore'` in pd.to_datetime, which is being deprecated.

### Changes:
- Wrapped call to df.to_datetime in a function and in a try: except: block
- Maintained the same results while removing the warning

### Testing:
- Confirmed the warning no longer appears
- Verified that date parsing still works correctly with both valid and invalid dates

### Original warning:
```
pybaseball/datahelpers/postprocessing.py:59: FutureWarning: errors='ignore' is deprecated and will raise in a future version. Use to_datetime without passing `errors` and catch exceptions explicitly instead
  data_copy[column] = data_copy[column].apply(pd.to_datetime, errors='ignore', format=date_format)
```

### DIFF
--- a/pybaseball/datahelpers/postprocessing.py
+++ b/pybaseball/datahelpers/postprocessing.py
@@ -56,7 +56,7 @@ def try_parse_dataframe(
             # the whole dataframe just tries to gobble up ints/floats as timestamps
             for date_regex, date_format in date_formats:
                 if isinstance(first_value, str) and date_regex.match(first_value):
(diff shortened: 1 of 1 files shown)


hsum:
This fixes issue #467 by removing the warning.

agene0001:
Hi I was wondering if this pr will be merged soon as I get the errors in my implementation of pybaseball still? It doesn't seem like that hard of a fix but I get it if for some reason it might be failing some tests.

